PR #2, opened 2016-03-01
Phase 1: Features from the current ATMO site

Uberi:
This is an initial PR to keep track of things and show current progress.

screenshot from 2016-03-01 18-08-39

* Provisioning clusters and workers is functional.
* Basic user management/login using Persona.
* New login pages, dashboard, etc.

There's a bug in django-browserid that prevents the app from running out of the box; there is a fix pending in https://github.com/mozilla/django-browserid/pull/298, but until then the tests won't pass.


vitillo:
@mreid-moz are we planning to support worker nodes? I thought we decided to drop them.

mreid-moz:
@vitillo I think we should support existing functionality for the first round - I wouldn't want to block the new service on having to make extensive changes to existing scheduled jobs.

vitillo:
@mreid-moz deal

analysis_service/base/templates/analysis_service/dashboard.jinja
```diff
@@ -0,0 +1,124 @@
...
+          <img width="48" src="{{ static("img/schedule.png") }}" />
+          Schedule a Spark analysis job
+        </a>
+      </p>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-6">
+      <h2>Launch a Worker</h2>
+      <p>Launch a server in the cloud and use it for custom map-reduce data analysis. The machine will be available for 24 hours, then it will be automatically terminated.</p>
```
mreid-moz: s/map-reduce//

analysis_service/base/forms.py
```diff
@@ -0,0 +1,448 @@
+import logging
+from django import forms
+from django.core.exceptions import ValidationError
+from django.template.defaultfilters import filesizeformat
+from analysis_service.base import models
+
+
+logger = logging.getLogger("django")
```
maurodoglio: I see the logger is not used anywhere in this module, can you please remove? Also, if you want to use it it's better to initialize it using 
```python
logger = logging.getLogger(__name__)
```
See [the python docs](https://docs.python.org/2/library/logging.html#logger-objects) for more info.

analysis_service/base/forms.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+class PublicKeyFileField(forms.FileField):
```
maurodoglio: Please add some docstrings here

analysis_service/base/forms.py
```diff
@@ -0,0 +1,448 @@
...
+    )
+    public_key = PublicKeyFileField(
+        required=True,
+        widget=forms.FileInput(attrs={'required': 'required'})
+    )
+
+    def save(self, user):
+        # create the model without committing, since we haven't
+        # set the required created_by field yet
+        new_cluster = super(NewClusterForm, self).save(commit=False)
```
maurodoglio: This is a bit confusing, I think you can make it simpler by overriding the __init__ method of NewClusterForm:
```python
def __init__(self, user, *args, **kwargs):
    self.created_by = user
    super(NewClusterForm, self).__init__(*args, **kwargs)
```
and then you can pass request.user directly to the NewClusterForm constructor.

analysis_service/base/forms.py
```diff
@@ -0,0 +1,448 @@
...
+            )
+        contents = uploaded_file.read()
+        if not contents.startswith('ssh-rsa AAAAB3'):
+            raise ValidationError(
+                'Invalid public key (a public key should start with \'ssh-rsa AAAAB3\')'
+            )
+        return contents
+
+
+class ClusterIdField(forms.IntegerField):
```
maurodoglio: I think you can use a [ModelChoiceField](https://docs.djangoproject.com/en/1.9/ref/forms/fields/#modelchoicefield) for both this field and ScheduledSparkIdField.

analysis_service/base/models.py
```diff
@@ -0,0 +1,206 @@
...
+class Worker(models.Model):
+    identifier = models.CharField(max_length=100)
+    public_key = models.CharField(max_length=100000)
+    start_date = models.DateTimeField(blank=True, null=True)
+    end_date = models.DateTimeField(blank=True, null=True)
+    created_by = models.ForeignKey(User, related_name='worker_created_by')
+    instance_id = models.CharField(max_length=50, blank=True, null=True)
+
+    def __str__(self):
+        return "<Worker {}>".format(self.identifier)
```
maurodoglio: __str__ should return a human-readable version  of the object. In this case it could simply return self.identifier.

static/.gitkeep
```diff
@@ -0,0 +1 @@
+Placeholder file to ensure that the static/ folder is kept by Git.
```
maurodoglio: You don't need to put the root static folder under version control. If the folder doesn't exist it will be created the first time collectstatic is run.

maurodoglio:
There are various files which are missing a blank line at the end. Can you please fix it?

analysis_service/base/models.py
```diff
@@ -0,0 +1,206 @@
+from datetime import datetime, timedelta
+from pytz import UTC
+from django.db import models
+from django.contrib.auth.models import User
+
+from analysis_service.base.util import provisioning, scheduling
+
+
+class Cluster(models.Model):
```
maurodoglio: Can you please add a description for the non-obvious fields using `help_text`?

maurodoglio: It would be great if you could identify eventual unique and unique-together keys so that we can enforce them as db constraints

Uberi: I don't think we have an explicit primary key for these, since you can currently just create two clusters with the same name. Do you think we should enforce that, for example, the identifiers must be unique for clusters?

Uberi:
Changes:

* Implement feedback re: models and forms.
* Clean up code a little bit.
* Fix scheduling stuff, left a debug statement in there by accident.

analysis_service/base/models.py
mreid-moz: Can we use the `jobflow_id`? Or do we need it before we've launched the cluster?

Uberi: That could potentially work for clusters, but not for scheduled jobs, which aren't associated with any particular jobflow. Might make testing a bit messier, though.

maurodoglio: > Do you think we should enforce that, for example, the identifiers must be unique for clusters?
I think we should either s/identifier/name or make the constraint more explicit. What is the identifier string used for?

maurodoglio:
I think you could still reduce the amount of code necessary for the form handling by using [class-based views](https://docs.djangoproject.com/en/1.9/topics/class-based-views/generic-editing/#form-handling-with-class-based-views), but let's keep that eventually for a followup patch.

maurodoglio:
@mreid-moz, @vitillo could anyone of you please briefly check the provisioning and scheduling code? I'm not very familiar with that yet. After that I think we can merge this. Thanks @Uberi !

analysis_service/base/models.py
mreid-moz: AFAIK it's not used for anything, it just tags the resources with the name so we can match up the instance or cluster with a job or a person in the AWS Console. It's convenient for housekeeping / debugging.

It should be about the same in ATMO - the name would appear in your list so you have some indication of what a given cluster / instance was doing.

analysis_service/base/templates/analysis_service/boot-script.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
vitillo: @mreid-moz can you review this?

analysis_service/settings.py
```diff
@@ -67,10 +69,42 @@
...
+    'WORKER_AMI':             'ami-0057b733',  # -> telemetry-worker-hvm-20151019 (Ubuntu 15.04)
+    'WORKER_PRIVATE_PROFILE': 'telemetry-example-profile',
+    'WORKER_PUBLIC_PROFILE':  'telemetry-example-profile',
+
+    # EMR configuration
+    # Master and slave instance types should be the same as the telemetry
+    # setup bootstrap action depends on it to autotune the cluster.
+    'MASTER_INSTANCE_TYPE':   'c3.4xlarge',
+    'SLAVE_INSTANCE_TYPE':    'c3.4xlarge',
+    'EMR_RELEASE':            'emr-4.2.0',
```
vitillo: This should be `emr-4.3.0`.

analysis_service/settings.py
```diff
@@ -67,10 +69,42 @@
...
+
+    # Make sure the ephemeral map matches the instance type above.
+    'EPHEMERAL_MAP':    {"/dev/xvdb": "ephemeral0", "/dev/xvdc": "ephemeral1"},
+    'SECURITY_GROUPS':  [],
+    'INSTANCE_PROFILE': 'telemetry-analysis-profile',
+    'INSTANCE_APP_TAG': 'telemetry-analysis-worker-instance',
+    'EMAIL_SOURCE':     'telemetry-alerts@mozilla.com',
+
+    # Buckets for storing S3 data
+    'TEMPORARY_BUCKET':    'bucket-for-ssh-keys',
```
vitillo: Does this bucket exist?

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+        "state":      cluster['Status']['State'],
+        "public_dns": cluster['MasterPublicDnsName'],
+    }
+
+
+def cluster_stop(jobflow_id):
+    emr.terminate_job_flows(JobFlowIds=[jobflow_id])
+
+
+def worker_start(user_email, identifier, public_key):
```
vitillo: @mreid-moz can you review this?

vitillo:
The provisioning and scheduling of Spark clusters looks OK to me.

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+        Instances={
+            'MasterInstanceType': settings.AWS_CONFIG['INSTANCE_TYPE'],
+            'SlaveInstanceType': settings.AWS_CONFIG['INSTANCE_TYPE'],
+            'InstanceCount': num_instances,
+            'Ec2KeyName': 'mozilla_vitillo',
+            'KeepJobFlowAliveWhenNoSteps': True,  # same as no-auto-terminate
+        },
+        JobFlowRole=settings.AWS_CONFIG["SPARK_INSTANCE_PROFILE"],
+        ServiceRole='EMR_DefaultRole',
+        Applications=[{'Name': 'Spark'}],
```
vitillo: Hive has to be added as well.

analysis_service/settings.py
Uberi: This is apparently where the service currently stores the keys ([1], [2]). I've previously tested that this particular code can successfully spawn workers, so the bucket probably exists now.

[1]: https://github.com/mozilla/telemetry-server/blob/f85fc6bb7920a524a3aa5b869140cf1641dd7873/http/analysis-service/server.py#L37
[2]: https://github.com/mozilla/telemetry-server/blob/f85fc6bb7920a524a3aa5b869140cf1641dd7873/http/analysis-service/config.py#L27

Uberi:
Changes:

* EMR version bumped to 4.3.0.
* Add Hive as well in the applications.

analysis_service/settings.py
vitillo: @mreid-moz I don't see this bucket from the console, do you know anything about it?

mreid-moz: This bucket does not exist.  On atmo, it gets replaced during deployment with the [actual bucket as a parameter](https://github.com/mozilla/telemetry-server/blob/master/http/analysis-service/analysis-service-stack.yaml#L65).

maurodoglio:
I think we are only missing a green light from @mreid-moz here. Mark any chance you can have a look at the provisioning code soon? 🙇

mreid-moz:
Yep, I'll take a look today.

Dockerfile
```diff
@@ -24,6 +24,9 @@ RUN ./bin/pipstrap.py
 COPY requirements.txt /app/requirements.txt
 RUN pip install --require-hashes --no-cache-dir -r requirements.txt
 
+# TODO: debug
```
mreid-moz: Can you file an issue for this TODO?

analysis_service/base/templates/analysis_service/boot-script.sh
mreid-moz: This file hasn't really changed from the [previous version](https://github.com/mozilla/telemetry-server/blob/master/http/analysis-service/templates/boot-script.sh) so it should be good.

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+def cluster_stop(jobflow_id):
+    emr.terminate_job_flows(JobFlowIds=[jobflow_id])
+
+
+def worker_start(user_email, identifier, public_key):
+    # upload the public key to S3
+    token = str(uuid4())
+    s3_key = 'keys/{}.pub'.format(token)
+    s3.put_object(
+        Bucket = settings.AWS_CONFIG['CODE_BUCKET'],
```
mreid-moz: If you upload the ssh key to the `CODE_BUCKET` bucket, you should fetch it from there too (not from `TEMPORARY_BUCKET` per line 97.

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+        Key = s3_key,
+        Body = BytesIO(public_key)
+    )
+
+    # generate the boot script for the worker
+    ephemeral_map = settings.AWS_CONFIG['EPHEMERAL_MAP']
+    boot_script = render_to_string('boot-script.sh', context={
+        'aws_region':       settings.AWS_CONFIG['AWS_REGION'],
+        'temporary_bucket': settings.AWS_CONFIG['TEMPORARY_BUCKET'],
```
mreid-moz: Per above, this bucket should match the location to which the key was uploaded above.

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+    token = str(uuid4())
+    s3_key = 'keys/{}.pub'.format(token)
+    s3.put_object(
+        Bucket = settings.AWS_CONFIG['CODE_BUCKET'],
+        Key = s3_key,
+        Body = BytesIO(public_key)
+    )
+
+    # generate the boot script for the worker
+    ephemeral_map = settings.AWS_CONFIG['EPHEMERAL_MAP']
```
mreid-moz: `EPHEMERAL_MAP` is an optional setting, is `settings` resilient to fetching missing keys? If it behaves like a regular `dict` (throwing a KeyError), this should change to `settings.AWS_CONFIG.get('EPHEMERAL_MAP')`

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+
+    return instance_id
+
+
+def worker_info(instance_id):
+    worker = ec2.describe_instances(
+        DryRun=False,
+        InstanceIds=[instance_id]
+    )['Reservations'][0]["Instances"][0]
```
mreid-moz: nit: use consistent quoting `'Instances'`

analysis_service/base/util/provisioning.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def worker_stop(instance_id):
+    ec2.terminate_instances(
+        DryRun=False,
+        InstanceIds=[instance_id]
+    )
+
+
+def get_tag_value(tags, key):
```
mreid-moz: This function is duplicated in `scheduling.py` - can we put it in a single place?

analysis_service/settings.py
mreid-moz: Looks like this code was storing keys in the "CODE" bucket, that's probably why it worked despite the missing bucket.

setup.py
```diff
@@ -7,6 +7,6 @@
     version='0.1dev',
     description='This is https://github.com/mozilla/analysis_service',
     author='Mozilla Foundation',
-    author_email='',
+    author_email='telemetry-analysis-service@mozilla.org',
```
mreid-moz: Does this email address exist?

mreid-moz:
Also, when I try to run the tests with `docker-compose run web ./manage.py test`, all the tests fail with the following error:
`ValueError: The file 'favicon.ico' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x7fd9d5946bd0>.`

setup.py
vitillo: Not that I am aware of.

Uberi:
Regarding the error you got, make sure to run `manage.py collectstatic` before the first run! I added a note to the README about that.

Changes:

* Use the code bucket for public keys (`TEMPORARY_BUCKET` is no longer used, so it's gone now).
* Make ephemeral map optional.
* Style fixes, such as double vs. single quotes, remove some debugging code (I originally used this to set things up in docker, but now that the README has a troubleshooting guide this is no longer needed).

jezdez:
Is there anything else needed here that I can help with?

maurodoglio:
I think we can merge this and eventually address remaining issues in another PR. @vitillo @mreid-moz are you ok with that?

mreid-moz:
@maurodoglio Sounds good to me.

vitillo:
@maurodoglio sounds good

maurodoglio:
I squashed and merged this PR manually.

PR #4, opened 2016-05-30
Add monitoring functionality

Uberi:
Based on #1.

With AWS Cloudwatch, we can do things like checking cluster health, job status, and so on.

This PR adds initial monitoring, and a small feature that notifies users if their cluster has been idle for an hour. When I'm running a task and switch to working on something else, if that fails, I get an email about it and can fix it right away.

Possible next steps with this monitoring capability include:

* Adding job stats to the dashboard directly, like "Number of jobs running", "Number of jobs completed", and "Average worker task load" to give a quick idea of cluster health.
* Ability to backup and kill idle nodes, potentially allowing us to get rid of cluster timeouts?


maurodoglio:
I'm gonna close this because I'm not sure what the actual changeset is. Happy to reopen (or cherrypick) once that's more clear.

PR #5, opened 2016-06-29
Various changes to the dev environment setup

maurodoglio:
See individual commits for details

### DIFF
--- a/analysis_service/settings.py
+++ b/analysis_service/settings.py
@@ -47,7 +47,6 @@ INSTALLED_APPS = [
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'django_browserid',
-    "sslserver"
 ]
 
 for app in config('EXTRA_APPS', default='', cast=Csv()):
--- a/analysis_service/settings.py
+++ b/analysis_service/settings.py
@@ -37,6 +37,7 @@ INSTALLED_APPS = [
 
     # Third party apps
     'django_jinja',
+    'whitenoise.runserver_nostatic',
 
     # Django apps
     'django.contrib.admin',
--- a/bin/run-dev.sh
+++ b/bin/run-dev.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 ./bin/run-common.sh
-./manage.py runsslserver 0.0.0.0:8000
+./manage.py runserver 0.0.0.0:8000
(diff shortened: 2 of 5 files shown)


analysis_service/base/migrations/0001_initial.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
jezdez: Before we set the generic name `base` into stone using the migration system, I'd strongly suggest to split the app into multiple apps `clusters`, `workers`, `spark_jobs` and a have the common code in the `analysis_service` package. In my experience having every domain objects in one package will eventually make maintenance much harder and adding/removing features risky. I'm happy to work on a separate PR after this got merged if you're okay with that pattern.

analysis_service/settings.py
```diff
@@ -37,6 +37,7 @@
 
     # Third party apps
     'django_jinja',
+    'whitenoise.runserver_nostatic',
```
jezdez: I think the wsgi file should also drop the use of the separate WSGI app, since when using the whitenoise middleware that's not needed anymore.

bin/run-dev.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 ./bin/run-common.sh
-./manage.py runsslserver 0.0.0.0:8000
+./manage.py runserver 0.0.0.0:8000
```
jezdez: IIRC Persona requires using SSL to login, have you confirmed Persona still working after this change?

jezdez:
LGTM with comments, rwc+

analysis_service/base/migrations/0001_initial.py
maurodoglio: Sounds like a very good idea, I'll drop this commit

bin/run-dev.sh
maurodoglio: It seems to be working even without SSL in the docker-compose setup

jezdez: Ah, you know what, I think I just misremembered that from MDN, we used a custom hostname, and I think Persona via `localhost` works without SSL. Sorry for the noise :)

maurodoglio: 👍

analysis_service/base/migrations/0001_initial.py
maurodoglio: I created [bug 1283187](https://bugzilla.mozilla.org/show_bug.cgi?id=1283187) to cover the followup work required.

analysis_service/settings.py
maurodoglio: @jezdez I updated the wsgi file, do you wanna have a quick look at it?

jezdez: Yep, r+

jezdez:
r+

maurodoglio:
Thanks!

PR #10, opened 2016-08-19
Bug 1296573 - fix form urls in jinja templates

maurodoglio:
Some of the forms were using outdated static urls.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/10)


maurodoglio:
r=me

PR #11, opened 2016-09-07
Set "VisibleToAllUsers" to True in cluster_start

robotblake:
When this flag is not set to True and using IAM roles, the clusters (as far as I can tell) become orphaned and are not visible or able to be terminated by anyone except the AWS account holder.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/11)

### DIFF
--- a/atmo/utils/provisioning.py
+++ b/atmo/utils/provisioning.py
@@ -51,7 +51,8 @@ def cluster_start(user_email, identifier, size, public_key):
             {'Key': 'Owner', 'Value': user_email},
             {'Key': 'Name', 'Value': identifier},
             {'Key': 'Application', 'Value': settings.AWS_CONFIG['INSTANCE_APP_TAG']},
-        ]
+        ],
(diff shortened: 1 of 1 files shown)


autolander:

maurodoglio:
Thanks for the patch. I'm still not 100% sure why this is needed, but I guess we can give it a try. Could you please copy the description of this pull request to the commit extended description?

robotblake:
"Whether the job flow is visible to all IAM users of the AWS account associated with the job flow. If this value is set to true, all IAM users of that AWS account can view and (if they have the proper policy permissions set) manage the job flow. If it is set to false, only the IAM user that created the job flow can view and manage it."

http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_RunJobFlow.html#EMR-RunJobFlow-request-VisibleToAllUsers

maurodoglio:
Thanks!

PR #12, opened 2016-09-13
Bug 1302495 - Update cluster status on page load

maurodoglio:
This is probably not the best place where to update the clusters status,
but until we have a async/periodic tasks system it's good enough.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/12)


maurodoglio:
Please ignore this pull request for now, we need to take a few decisions before we can actually fix this problem

PR #13, opened 2016-09-15
Bug 1301048 - Setup RQ worker and scheduler

maurodoglio:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/13)

### DIFF
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -46,6 +46,7 @@ INSTALLED_APPS = [
     # Third party apps
     'django_jinja',
     'whitenoise.runserver_nostatic',
+    'django_rq',
 
     # Django apps
     'django.contrib.admin',
@@ -123,6 +124,16 @@ REDIS_URL = config(
     'REDIS_URL',
     cast=parse.urlparse
 )
+
+RQ_QUEUES = {
+    'default': {
+        'HOST': REDIS_URL.hostname,
+        'PORT': REDIS_URL.port,
+        'DB': 0,
+        'PASSWORD': REDIS_URL.password,
+        'DEFAULT_TIMEOUT': 600,
+    }
+}
 
 # Add the django_browserid authentication backend.
 AUTHENTICATION_BACKENDS = (
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -10,6 +10,11 @@ https://docs.djangoproject.com/en/1.9/ref/settings/
 
 import os
 
+try:
+    import urlparse as parse
+except ImportError:
+    from urllib import parse
+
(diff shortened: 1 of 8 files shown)


atmo/settings.py
```diff
@@ -10,6 +10,11 @@
 
 import os
 
+try:
+    import urlparse as parse
+except:
```
jezdez: Mind using `except ImportError` here?

atmo/settings.py
```diff
@@ -114,6 +120,21 @@
...
+REDIS_URL = config(
+    'REDIS_URL',
+    cast=parse.urlparse
+)
+
+RQ_QUEUES = {
+    'default': {
+        'HOST': REDIS_URL.hostname,
+        'PORT': REDIS_URL.port,
+        'DB': 0,
```
jezdez: Just in theory, in case we ever require to specify a specific db, we could use the path component of the `REDIS_URL` value. I think it's okay to stick to 0 for now

atmo/settings.py
```diff
@@ -114,6 +120,21 @@
...
+        'PASSWORD': REDIS_URL.password,
+        'DEFAULT_TIMEOUT': 360,
```
jezdez: That means jobs will be considered lost if they don't finish in 6 minutes, do you think that's okay? Should we increase to 10 mins?

atmo/urls.py
```diff
@@ -17,5 +17,6 @@
     # contribute.json url
     url(r'^(?P<path>contribute\.json)$', static.serve, {'document_root': settings.ROOT}),
 
+    url(r'^django-rq/', include('django_rq.urls')),
```
jezdez: The dashboard [uses the Django admin templates](https://github.com/ui/django-rq/blob/3d0091b362152bbacb23a3a810572eb72f574d9f/django_rq/templates/django_rq/stats.html#L1), so let's mount that under `admin/rq/` instead of `django-rq`

jezdez:
In my experience with Celery it was often useful to have queues for different levels of priority, e.g. have long and short running jobs, so that there is not a long running job blocking a shorter one. Do you think that'll be needed for atmo at the moment?

atmo/settings.py
maurodoglio: I don't mind changing it, we can still tweak it later when we have a better idea of the typical task duration

maurodoglio:
I don't think we are gonna need queues with different priorities soon, let's keep this for later.

atmo/settings.py
jezdez: Works for me

PR #14, opened 2016-09-15
Bug 1301037 - Add option to choose EMR release

maurodoglio:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/14)


atmo/clusters/forms.py
```diff
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
```
jezdez: Don't think that's needed

atmo/clusters/models.py
```diff
@@ -36,6 +40,13 @@ class Cluster(models.Model):
         max_length=50, blank=True, null=True,
         help_text="AWS cluster/jobflow ID for the cluster, used for cluster management."
     )
+
+    emr_release = models.CharField(
+        max_length=50, choices=zip(*(EMR_RELEASES,)*2), default=EMR_RELEASES[-1],
```
jezdez: `zip` returns a `zip` object on Python 3, not sure if the choices parameter handles that nicely. wrapping it in a `list` would appease my mind.

Side note: did we ever want to make atmo Python 3 btw?

atmo/clusters/models.py
```diff
@@ -7,7 +7,11 @@
 from ..utils import provisioning
 
 
+EMR_RELEASES = ('5.0.0', '4.5.0')
```
jezdez: Maybe add a comment here that the last one will be the default?

atmo/templates/atmo/new-cluster.jinja
```diff
@@ -33,6 +33,13 @@
           <p class="help-block"></p>
         </div>
         <div class="alert alert-warning">Upload your SSH <strong>public key</strong>, not private key! This will generally be found in places like <code>~/.ssh/id_rsa.pub</code>.</div>
+        <div class="form-group">
+          <label>EMR release</label>
+          <div class="control-group">
+            {{ new_cluster_form.emr_release }}
```
jezdez: I only now realize we don't print any form errors in that template, any reason not to?

atmo/clusters/models.py
maurodoglio: We have never discussed that tbh. Let's have a chat about it.

atmo/templates/atmo/new-cluster.jinja
maurodoglio: I would rewrite the whole form template tbh. There are a few things there that break the DRY principle at the moment. Let's talk about it.

atmo/clusters/models.py
maurodoglio: Good idea

maurodoglio:
This pull request has already been merged

PR #17, opened 2016-09-20
Bug 1303104 - Use forked version of rq-scheduler

maurodoglio:
My fork of rq-scheduler supports running multiple schedulers
side-by-side. Once my patch gets merged we can point the requirements
back to the upstream version.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/17)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,5 +29,8 @@ rq_scheduler:
   links:
     - db
     - redis
+    # django-rq doesn't support rqscheduler `retry` mode yet
+    # so we need to use the original startup script, which retrieves
+    # the
   command:
-    python manage.py rqscheduler
+    rqscheduler --url=redis://redis:6379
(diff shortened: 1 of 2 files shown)


maurodoglio:
@jezdez r?

docker-compose.yml
```diff
@@ -29,5 +29,8 @@ rq_scheduler:
   links:
     - db
     - redis
+    # django-rq doesn't support rqscheduler `retry` mode yet
+    # so we need to use the original startup script, which retrieves
+    # the
   command:
-    python manage.py rqscheduler
+    rqscheduler --url=redis://redis:6379
```
jezdez: I wonder if we should simply set `RQ_REDIS_URL` as well and remove `--url ..`, to have that config value in one place instead of many? https://github.com/ui/rq-scheduler/blob/396efadda8610548b474e680507b278676fc2262/rq_scheduler/scripts/rqscheduler.py#L21

jezdez:
LGTM r+

docker-compose.yml
maurodoglio: I would agree if this was a long term thing, but at the end it's gonna be there (hopefully) for a short time, until the patch gets merged upstream and the new parameter is added to django-rq as well.

PR #20, opened 2016-09-27
Bug 1304972 - Get rid of django-crontab in favor of rq jobs

maurodoglio:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/20)


atmo/clusters/tests.py
```diff
@@ -12,7 +12,7 @@
 class TestCreateCluster(TestCase):
     @mock.patch('atmo.utils.provisioning.cluster_start', return_value=u'12345')
     def setUp(self, cluster_start):
-        self.start_date = datetime.now().replace(tzinfo=UTC)
+        self.start_date = timezone.now()
```
jezdez: 💥

atmo/workers/management/commands/delete_workers.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Delete expired workers'
+
+    def handle(self, *args, **options):
+        now = timezone.now()
+
+        for worker in Worker.objects.all():
+            if worker.end_date >= now:  # the worker is expired
```
jezdez: Let's use `Worker.objects.filter(end_date__gte=now)` here instead

atmo/clusters/models.py
```diff
@@ -94,16 +102,18 @@ def save(self, *args, **kwargs):
 
         # set the dates
         if not self.start_date:
-            self.start_date = datetime.now().replace(tzinfo=UTC)
+            self.start_date = timezone.now()
```
jezdez: Since `timezone.now()` will have microsecond accuracy it's a minor nit to call it again to set `end_date`. Instead set it into a separate variable and add the timedelta to that for `end_date`

atmo/clusters/tasks.py
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
jezdez: The terminology for RQ is jobs and not tasks, mind renaming the file to `jobs.py`?

atmo/clusters/tasks.py
```diff
@@ -0,0 +1,12 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+from django.core.management import call_command
+
+
+def delete_clusters():
```
jezdez: From an API point of view, I'd rather see the app logic of the jobs to live here and the management commands to call the Python functions defined here -- both to have a consistent place for the various jobs (and not have to look for the command files) and also to not have `call_command` in the call stack at all

atmo/schedule.py
```diff
@@ -0,0 +1,41 @@
...
+        'timeout': 5
+    },
+}
+
+
+def register_job_schedule():
+    scheduler = django_rq.get_scheduler()
+    for job_id, params in job_schedule.items():
+        scheduler.cron(params['cron_string'], id=job_id,
+                       func=params['func'], timeout=params.get('timeout'))
```
jezdez: Let's define a default value for the `timeout` parameter here, that we agree out jobs should be running at the longest. The default it falls back to is 180 seconds

atmo/workers/management/commands/delete_workers.py
maurodoglio: 👍

atmo/clusters/tests.py
maurodoglio: 👻

atmo/clusters/models.py
maurodoglio: :+1

atmo/clusters/tasks.py
maurodoglio: 👍

atmo/clusters/tasks.py
maurodoglio: 👍

atmo/schedule.py
maurodoglio: :+1:

PR #28, opened 2016-10-06
Bug 1307528 - Backup log files to s3

fbertsch:
@jezdez r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/28)

### DIFF
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -102,6 +102,7 @@ AWS_CONFIG = {
     'CODE_BUCKET': 'telemetry-analysis-code-2',
     'PUBLIC_DATA_BUCKET': 'telemetry-public-analysis-2',
     'PRIVATE_DATA_BUCKET': 'telemetry-private-analysis-2',
+    'LOG_BUCKET':          'telemetry-analysis-logs-2'
 }
 for aws_cred in ('AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY', 'AWS_DEFAULT_REGION'):
     if aws_cred not in os.environ:
--- a/atmo/utils/provisioning.py
+++ b/atmo/utils/provisioning.py
@@ -1,6 +1,7 @@
 from uuid import uuid4
 
 from django.conf import settings
+from datetime import datetime
 import boto3
 import requests
 
@@ -25,8 +26,13 @@ def cluster_start(user_email, identifier, size, public_key, emr_release):
             settings.AWS_CONFIG['SPARK_EMR_BUCKET']
         )
     ).json()
+
+    now = datetime.utcnow().isoformat()
+    log_uri = 's3://{}/{}/{}'.format(settings.AWS_CONFIG['LOG_BUCKET'], identifier, now)
+
     cluster = emr.run_job_flow(
         Name=str(uuid4()),
+        LogUri=log_uri,
         ReleaseLabel='emr-{}'.format(emr_release),
         Instances={
             'MasterInstanceType': settings.AWS_CONFIG['INSTANCE_TYPE'],
--- a/atmo/utils/scheduling.py
+++ b/atmo/utils/scheduling.py
@@ -1,6 +1,7 @@
 from uuid import uuid4
 
 from django.conf import settings
+from datetime import datetime
 import boto3
 import requests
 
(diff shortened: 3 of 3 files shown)


jezdez:
@fbertsch Apologies for the slow response, we're working on getting atmo2 running on Heroku right now and will get back to this asap, probably next week

atmo/settings.py
```diff
@@ -107,6 +107,7 @@
     'CODE_BUCKET':         'telemetry-analysis-code-2',
     'PUBLIC_DATA_BUCKET':  'telemetry-public-analysis-2',
     'PRIVATE_DATA_BUCKET': 'telemetry-private-analysis-2',
+    'LOG_BUCKET':          'telemetry-analysis-logs-2'
```
jezdez: Is that the bucket that has been used before as mentioned in the bug?

jezdez:
This feature was also added in https://github.com/mozilla/telemetry-server/pull/157 and uses a different log location, @mreid-moz can you elaborate which we should use for v2?

mreid-moz:
We used that particular bucket in the telemetry-server repo for two main reasons:
1. It is a bucket known to have write permission from the clusters, and is part of the overall provisioning for the atmo-related resources.
2. It could be relatively easily incorporated into the "view logs" part of scheduled jobs on atmo (though at the moment, that view doesn't handle subdirs, so I don't think it _actually_ works).

As long as both of those criteria are satisfied here, I think it's preferable to keep the cluster logs in a separate bucket.

> This looks technically correct, but I have questions about how to handle deleted jobs/clusters. Do we need to purge the log dirs, too?

If we proceed with a log-specific bucket, I would recommend setting a bucket expiration policy to automatically remove logs after some time (30 or 60 days, say).

One last comment - it might be a good idea to include a date component in the log URI somewhere so that logs can be differentiated by cluster launch time. A long list of cluster ids is not the most convenient for finding recent info.

fbertsch:
@mreid-moz, @jezdez r?
Sorry about the delay. I mentioned this in the bug as well - wasn't sure if we wanted to go separate bucket or same log location as before. 

I went ahead and added the ISO time, so the same identifier will have multiple time buckets.

mreid-moz:
@fbertsch Should we put the timestamp before the identifier so we can look at the list sorted? or is this info going to be exposed thru the atmo UI?

fbertsch:
@mreid-moz If the timestamp is before the identifier I don't really see any reason to have the identifier at all, since we won't have any overlap in timestamps (i.e. it will always be bucket/one_timestamp/one_identifier/logs). What advantage would it bring to put the timestamp first?

Also I think we do want this expose through the web UI, if not initially then definitely later.

mreid-moz:
@fbertsch my thinking was that when looking at the logs with awscli or via the console, it would be more useful to see logs ordered by time than by cluster ids, since one is more likely to know the (rough) time a cluster did work rather than its id. 

Maybe I'm mistaken about what `identifier` is, though. Does it correspond to the _name_ of the cluster rather than its `j-123ABC456DEF7` ID?  If so, then putting the timestamp after the name as you've done makes sense to me.

In either case, if the logs will be exposed via the atmo UI in the near future, then adding the timestamp is probably unneccessary.

fbertsch:
@mreid-moz The identifier is user entered, and comes up as the 'Name' value in the tags, so it is helpful when looking over the litany of logs.

For the UI, I'm thinking that we *should* keep it as-is for the web UI. The reason being if we just do bucket/identifier, each time a job runs it will show up as bucket/identifier/cluster-ID (the j-123ABC), so the UI will have to figure out which one of those cluster-IDs to the job at hand, rather than knowing it's a single prefix before even starting up the cluster. @jezdez, do you agree?

Wow, this is a lot of talk for such a simple change :)

jezdez:
Two things to note:

1. The identifier can be changed by the user after the cluster is created, so logs may be distributed in multiple subfolders in S3 when using the identifier as the root folder.

2. [bug 1310157](https://bugzilla.mozilla.org/show_bug.cgi?id=1310157) describes the log viewer feature for both the Spark jobs as well as the clusters, so we'll definitely see it shown eventually in the frontend (although not in this release, so not part of this PR)

vitillo:
@jezdez I am not sure how changing the identifier of the cluster is useful; maybe we could remove that feature?

fbertsch:
@jezdez RE: The identifer can change,
If it can change I don't think we should use it here then. We probably could just put all the logs in a timestamped folder, which should include no overlap if there's second granularity.
(bucket/timestamp, and remove identifier entirely). The backend DB will need to store the associated timestamp for the job then.

jezdez:
@vitillo Sure, that feature was there when we picked up atmo2 from the intern, so I don't know if there was a need for this. We recently made the Spark job identifier unique in #47 that also added a identifier check in the frontend, so I think it makes sense to stick to it once chosen.

fbertsch:
@jezdez, @vitillo then we can leave the logs in their current location if identifier is not changing

maurodoglio:
Considering the release date is approaching I would suggest to do whatever atmo v1 is doing in terms of paths, buckets ,etc. We can still iterate on this after the prod release

jezdez:
@fbertsch ok, I'll pick this PR up on Monday with the changes to disable identifier editing

maurodoglio:
I took care of making the job/cluster identifier immutable in PR 58.

maurodoglio:
Closing this in favor of #59

PR #31, opened 2016-10-10
Bug 1308713 - Instrument RQ jobs

jezdez:
This means the RQ jobs are instrumented with NewRelic by using RQ's SimpleWorker class that doesn't fork a new process per job but does the job performing inline.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/31)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 web: newrelic-admin run-program gunicorn atmo.wsgi:application --workers 4 --log-file -
-worker: newrelic-admin run-program python manage.py rqworker default
+worker: newrelic-admin run-program python manage.py rqworker --worker-class=rq.SimpleWorker default
 # django-rq doesn't support rqscheduler retry mode yet
 # so we need to use the original startup script
 scheduler: newrelic-admin run-program rqscheduler --url=$REDIS_URL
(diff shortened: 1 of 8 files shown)


Procfile
```diff
@@ -1,5 +1,5 @@
 web: newrelic-admin run-program gunicorn atmo.wsgi:application --workers 4 --log-file -
-worker: newrelic-admin run-program python manage.py rqworker default
+worker: newrelic-admin run-program python manage.py rqworker --worker-class=rq.SimpleWorker default
```
maurodoglio: shall we use the same worker class for local development?

atmo/utils/middleware.py
```diff
@@ -0,0 +1,16 @@
...
+
+class NewRelicPapertrailMiddleware(object):
+
+    def __init__(self):
+        self.heroku_app_name = os.environ.get('HEROKU_APP_NAME')
+
+    def process_request(self, request):
+        if not self.heroku_app_name:
+            return
+        log_url = ("https://papertrailapp.com/systems/%s/events?time=#%d" %
```
jezdez: whoops, that `#` shouldn't be there

maurodoglio:
🚢 it!

PR #32, opened 2016-10-10
Populate the environment with AWS credentials

maurodoglio:
We need to do this because Boto doesn't know about our .env file.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/32)

### DIFF
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -108,6 +108,10 @@ AWS_CONFIG = {
     'PUBLIC_DATA_BUCKET':  'telemetry-public-analysis-2',
     'PRIVATE_DATA_BUCKET': 'telemetry-private-analysis-2',
 }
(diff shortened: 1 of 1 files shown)


autolander:

atmo/settings.py
```diff
@@ -108,6 +108,10 @@
     'PUBLIC_DATA_BUCKET':  'telemetry-public-analysis-2',
     'PRIVATE_DATA_BUCKET': 'telemetry-private-analysis-2',
 }
+for aws_cred in ('AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY'):
```
jezdez: Let's add `AWS_DEFAULT_REGION` as well

maurodoglio: 👍

atmo/settings.py
```diff
@@ -108,6 +108,10 @@
...
+    if not aws_cred in os.environ:
```
maurodoglio: 👍

PR #41, opened 2016-10-13
Bug 1309901 - Use spot instances for ad-hoc clusters

fbertsch:
@jezdez r?

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/41)

### DIFF
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -84,13 +84,13 @@ WSGI_APPLICATION = 'atmo.wsgi.application'
 AWS_CONFIG = {
     # AWS EC2 configuration
     'AWS_REGION': 'us-west-2',
-    'INSTANCE_TYPE': 'c3.4xlarge',
 
     # EMR configuration
     # Master and slave instance types should be the same as the telemetry
     # setup bootstrap action depends on it to autotune the cluster.
     'MASTER_INSTANCE_TYPE': 'c3.4xlarge',
-    'SLAVE_INSTANCE_TYPE': 'c3.4xlarge',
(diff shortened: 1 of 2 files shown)


autolander:

atmo/settings.py
```diff
@@ -91,6 +91,7 @@
     # setup bootstrap action depends on it to autotune the cluster.
     'MASTER_INSTANCE_TYPE': 'c3.4xlarge',
     'SLAVE_INSTANCE_TYPE': 'c3.4xlarge',
```
jezdez: Let's rename that setting to `WORKER_INSTANCE_TYPE` to math its use below

atmo/settings.py
```diff
@@ -91,6 +91,7 @@
...
+    'CORE_SPOT_BID': '0.84',
```
jezdez: If yes, wrapping this into a `config` call that allows us to set this via an environment variable (see other settings for an example) would be useful

fbertsch:
@jezdez made the changes you recommended, RFAL.

vitillo:
@fbertsch How has this been tested? How much longer does it take on average for a cluster to bootstrap now?

jezdez:
I have only done code review, not functional testing but you can try it on https://atmo-staging.herokuapp.com which is automatically deployed with master

jezdez:
FTR We missed on case of the old variable name, fixed in https://github.com/mozilla/telemetry-analysis-service/commit/732b2f75d1493581f8331c7f8a263ba6aa113cc5

PR #27, opened 2016-10-06
Move to Heroku

jezdez:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/27)

PR #48, opened 2016-10-20
Bug 1311634 - Add back circle.yml

robotblake:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/48)


autolander:

jezdez:
PR #27 removed the `CMD` statement from the `Dockerfile` since we wanted to reuse it for the various Python based services in `docker-compose.yml` (web, worker, scheduler) (https://github.com/mozilla/telemetry-analysis-service/commit/684d1c81cced5f3482a715fb4fab815e6cd09963#diff-3254677a7917c6c01f55212f86c57fbf).

That means we'll have to add a new `bin/run_web_prod` with the content:

```
#!/usr/bin/env bash
./manage.py migrate
newrelic-admin run-program gunicorn atmo.wsgi:application -b 0.0.0.0:${PORT:-8000} --log-file -
```

..and add it to the `Dockerfile` with `CMD ["./bin/run_web_prod"]`.

About the other services `worker` and `scheduler` we'll need to have something similar right? 

How are the services defined in the dockerflow? Will it just use docker-compoase as well and we can define a docker-compose-prod.yml or something else?

Note: Since we've moved away from using NewRelic's Python based initialization and instead use `newrelic-admin run-program` to prepend any command to be monitored by it. Newrelic requires setting various env vars for this to work correctly. See the Heroku `Procfile` for all the commands and the Heroku project for the env vars needed.

jezdez:
I've opened [bug 1311634](https://bugzilla.mozilla.org/show_bug.cgi?id=1311634) to track this work

robotblake:
We specify the processes to run in our deploy repo, I already copied over what was in the Procfile. You don't _need_ to specify a CMD in the Dockerfile, though I'm wondering if writing a generalized `bin/run` script that can take the argument of `web`, `worker`, or `scheduler` might make sense.

robotblake:
I added a script that does what I mentioned above.

jezdez:
I've made some additional updates like moving all tests into using Docker and getting rid of tox. The rest looked 👍 Thanks @robotblake!

jezdez:
@robotblake Oh before I forget, in case the deploy setup needs to be updated, I had to revert using the Docker entrypoint (in 06410510050cf428277476bc0651a96a793f0aa3) since we use `docker-compose run` in the Makefile to run the tests and other tasks, and it was complaining that `bin/run` wouldn't accept the parameters (e.g. `docker-compose run web ./manage.py collectstatic` would mean it'd call `bin/run ./manage.py collectstatic`.

PR #53, opened 2016-10-21
Bug 1311457 - Show termination time in cluster detail page instead of timedelta

jezdez:
This also shows a tooltip for the timedelta and the current UTC time as a tooltip over the time in the footer

### DIFF
--- a/atmo/static/js/refresher.js
+++ b/atmo/static/js/refresher.js
@@ -5,8 +5,17 @@ $(function() {
           refresh_container = '#refresh-container',
           refresh_timer = '#refresh-timer',
           refresher = $('#refresher'),
+          time = $('#time'),
           timeout = 60,
           timeout_id;
+      var utc_now = function() {
+        return moment().utcOffset(0).format('YYYY-MM-DD HH:mm:ss');
+      }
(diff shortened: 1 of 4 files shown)


jezdez:
screen shot 2016-10-21 at 15 24 37
screen shot 2016-10-21 at 15 24 30

jezdez:
@maurodoglio r?

atmo/static/js/refresher.js
```diff
@@ -5,8 +5,17 @@ $(function() {
...
           refresh_timer = '#refresh-timer',
           refresher = $('#refresher'),
+          time = $('#time'),
           timeout = 60,
           timeout_id;
+      var utc_now = function() {
+        return moment().utcOffset(0).format('YYYY-MM-DD HH:mm:ss');
+      }
+      var updateTime = function() {
+        time.attr('data-original-title', utc_now());
```
maurodoglio: Where is data-original-title used? I feel like I'm missing something.

jezdez: That's the magic attribute title of the tooltip plugin: http://stackoverflow.com/a/27237034/153296

PR #57, opened 2016-10-25
Bug 1307087 - Use full notebook uri for scheduling

maurodoglio:
This also fixes an issue with the last_run_date non getting updated


maurodoglio:
@jezdez  r?

PR #58, opened 2016-10-26
Bug 1312983 - Make job identifier immutable

maurodoglio:


maurodoglio:
@jezdez r?

jezdez:
I'm not following, why do you want to remove editing job details?

maurodoglio:
I disabled editing the job identifier in the job edit form and I removed the cluster edit view completely because the only value that was edited there was the cluster identifier. This seems to have little value for the end users and we want to use the cluster and job identifiers as part of the path to their logs.

jezdez:
@maurodoglio Ah, right, makes sense.

PR #59, opened 2016-10-26
Bug 1307528 - Backup log files to s3

maurodoglio:
I updated @fbertsch code and rebased it on top of latest master.

### DIFF
--- a/atmo/provisioning.py
+++ b/atmo/provisioning.py
@@ -4,6 +4,7 @@
 from uuid import uuid4
 
 from django.conf import settings
+from django.utils import timezone
 import requests
 
 from .aws import emr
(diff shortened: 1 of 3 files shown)


autolander:

jezdez:
LGTM r+

jezdez:
Oh one thing, we may want to delete the logs of terminated clusters and deleted jobs to prevent overwrites with new clusters/jobs with he same identifier.

fbertsch:
@jezdez We don't want to overwrite since we'll need to use these logs to debug problems in the cluster, especially after shutdown. I don't think there will be overwrite since the folder includes the timestamp, do you agree? E.g. later created clusters with the same identifier will have a different log directory.

jezdez:
@fbertsch Yeah, timestamp is good, but keeping every log ever is not going to help finding the logs we need. @maurodoglio said there is a way to set a bucket level expiration policy

maurodoglio:
I set the bucket expiration policy to 61 days.

maurodoglio:
Thanks for the patch @fbertsch 😸

fbertsch:
Thanks for fixing it up guys 🎉

PR #60, opened 2016-10-26
Bug 1313008 - Don't delete expired jobs

maurodoglio:
No jobs should be deleted without their owner's consent.
This patch also fixes a couple of issues with the expiration logic,
namely:
 - the last_run_date vs timeout check was wrong
 - only RUNNING jobs should be terminated because of a timeout

### DIFF
--- a/atmo/jobs/models.py
+++ b/atmo/jobs/models.py
@@ -11,6 +11,7 @@ from django.utils.functional import cached_property
 
 from ..models import EMRReleaseModel
 from .. import provisioning, scheduling
+from ..clusters.models import Cluster
 
 
 class SparkJob(EMRReleaseModel):
@@ -100,13 +101,14 @@ class SparkJob(EMRReleaseModel):
         return self.most_recent_status
 
     def is_expired(self, at_time=None):
-        if self.current_run_jobflow_id is None:
-            return False  # job isn't even running at the moment
+        """Tells whether the current job run has run out of time or not"""
+        if not self.current_run_jobflow_id or not self.last_run_date:
+            # Job isn't even running at the moment and never ran before
+            return False
         if at_time is None:
             at_time = timezone.now()
-        if self.last_run_date and self.last_run_date + timedelta(hours=self.job_timeout) >= at_time:
-            return True  # current job run expired
-        return False
+        max_run_time = self.last_run_date + timedelta(hours=self.job_timeout)
+        return self.most_recent_status not in Cluster.FINAL_STATUS_LIST and at_time >= max_run_time
 
     def should_run(self, at_time=None):
         """Return True if the scheduled Spark job should run, False otherwise."""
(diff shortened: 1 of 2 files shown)


maurodoglio:
@jezdez  r?

atmo/jobs/models.py
```diff
@@ -100,12 +100,16 @@ def update_status(self):
...
+        """Tells whether the current job run has run out of time or not"""
+        if not self.current_run_jobflow_id or not self.last_run_date:
+            # Job isn't even running at the moment
+            return False
         if at_time is None:
             at_time = timezone.now()
-        if self.last_run_date and self.last_run_date + timedelta(hours=self.job_timeout) >= at_time:
-            return True  # current job run expired
+        max_run_time = self.last_run_date + timedelta(hours=self.job_timeout)
+        if self.most_recent_status == "RUNNING" and at_time >= max_run_time:
```
jezdez: I think this also may be "WAITING"? Should we turn around the statement and use `if self.most_recent_status not in Cluster.FINAL_STATUS_LIST and at_time >= max_run_time`?

atmo/jobs/models.py
```diff
@@ -100,12 +100,16 @@ def update_status(self):
...
+            # Current job run expired
+            return True
```
jezdez: We can simply do a return of the condition value instead of explicitly having those return statements here, if you're prefer

tests/test_jobs.py
```diff
@@ -246,3 +246,55 @@ def test_spark_job_second_run(now, test_user):
...
+    spark_job.current_run_jobflow_id = jobflow_id
+    spark_job.last_run_date = timeout_run_date + timedelta(seconds=1)
+    spark_job.most_recent_status = running_status
+    assert not spark_job.is_expired(at_time=now)
+
+    # All the conditions are met
+    spark_job.current_run_jobflow_id = jobflow_id
+    spark_job.last_run_date = timeout_run_date
+    spark_job.most_recent_status = running_status
+    assert spark_job.is_expired(at_time=now)
```
jezdez: A test is missing for not passing in `at_time`

atmo/jobs/models.py
```diff
@@ -100,12 +100,16 @@ def update_status(self):
         return self.most_recent_status
 
     def is_expired(self, at_time=None):
-        if self.current_run_jobflow_id is None:
-            return False  # job isn't even running at the moment
+        """Tells whether the current job run has run out of time or not"""
+        if not self.current_run_jobflow_id or not self.last_run_date:
+            # Job isn't even running at the moment
```
jezdez: ..`and never ran before`

jezdez:
Note: I've fixed the Circle CI issues in master

atmo/jobs/models.py
maurodoglio: 👍

atmo/jobs/models.py
maurodoglio: 👍

atmo/jobs/models.py
maurodoglio: 👍

PR #61, opened 2016-10-26
Bug 1313031 - Render Notebook content on refresh

jezdez:


jezdez:
@maurodoglio r?

atmo/static/js/jobs.js
```diff
@@ -1,13 +1,31 @@
 $(function() {
-    var root = this;
-    var $holder = $("#notebook-holder");
-    var content = $("#notebook-content").val();
+  $.fn.atmoNotebook = function(options) {
+    var settings = $.extend({
```
maurodoglio: `settings` doesn't seem to be used anywhere

atmo/views.py
```diff
@@ -37,3 +40,20 @@ def dashboard(request):
         'user_spark_jobs': jobs,
     }
     return render(request, 'atmo/dashboard.html', context=context)
+
+
+@requires_csrf_token
```
jezdez: The decorator is needed to handle 500 errors that use the CSRF token. The view is needed to pass the request to the template context of the 500.html since it's by default not using the context processors that would add that request to the context, and doesn't either in the [default implementation of the server error view](https://github.com/django/django/blob/9c956ff2f559e7074b5e49449a16857e9c47e3bd/django/views/defaults.py#L51-L63). In theory that's because the request itself may be the source of a server error, so Django doesn't do that by default. In practice though it's rarely an issue and the exception handling in modern Django can show the original exception just fine. Oh, and we need to request for a few template tags in the base template.

jezdez: The decorator is needed to handle 500 errors that use the CSRF token. The view is needed to pass the request to the template context of the 500.html since it's by default not using the context processors that would add that request to the context, and doesn't either in the default implementation of the server error view. In theory that's because the request itself may be the source of a server error, so Django doesn't do that by default. In practice though it's rarely an issue and the exception handling in modern Django can show the original exception just fine. Oh, and we need to request for a few template tags in the base template.

atmo/static/js/jobs.js
jezdez: Ah, true, removing it, this was a leftover from a previous iteration of that jQuery plugin

PR #65, opened 2016-10-28
Factor confirmation form into own mixin

jezdez:
This also uses django.forms.models.ModelForm as the
base class of the mixin to have a consistent bases for
our forms.


autolander:

jezdez:
@maurodoglio r?

jezdez:
r=me

PR #67, opened 2016-11-04
Bug 1307087 - Refactor job status tracking

jezdez:


atmo/jobs/models.py
```diff
@@ -98,20 +100,50 @@ def __str__(self):
...
+        return self.most_recent_status in self.FINAL_STATUS_LIST
+
+    @property
+    def is_runnable(self):
+        return self.has_never_run or self.has_finished
+
+    @property
+    def is_expired(self):
+        """Whether the current job run has run out of time"""
+        if not self.current_run_jobflow_id or self.last_run_date is None:
```
jezdez: @maurodoglio Here's still a chance of a race condition since we check `current_run_jobflow_id` and `last_run_date` to decide if a job is expired and not `most_recent_status` to decide if it's expired or not.

The first are stored by atmo when a job is run, the latter when we update the status from AWS. Which means I'm wondering if that's the right condition.

In the tests there is this:
```
    # A spark job cannot expire if:
    # it doesn't have a jobflow_id OR
    # it doesn't have a a last_run_date OR
    # its most_recent_status is not RUNNING OR
    # it hasn't run for more than its timeout
```

atmo/jobs/models.py
```diff
@@ -99,13 +101,32 @@ def __repr__(self):
...
+    @property
+    def is_runnable(self):
+        return self.has_never_run or self.has_finished
+
+    @property
     def is_expired(self):
-        """Tells whether the current job run has run out of time or not"""
-        if not self.current_run_jobflow_id or not self.last_run_date:
+        """Whether the current job run has run out of time"""
+        if not self.current_run_jobflow_id or self.last_run_date is None:
```
maurodoglio: You could use the new `has_never_run` property here and get rid of the check on `current_run_jobflow_id`. I think this condition at the moment only matches if the job has never run, because afaik we never reset those properties between jobs.

PR #70, opened 2016-11-08
Bug 1315670 - Reset a job schedule when start_date is changed

maurodoglio:


maurodoglio:
@jezdez could you r? please?

atmo/jobs/forms.py
```diff
@@ -153,12 +154,38 @@ class EditSparkJobForm(BaseSparkJobForm):
...
+            if self.cleaned_data['start_date'] > now:
+                # If the start_date changed and it's set in the future
+                # reset the last_run_date so that it runs at start_date.
+                self.instance.last_run_date = None
+                print 'setting to none'
+            else:
+                # If the start_date changed and it's set in the past
+                # set last_run_date to start_date, so that it runs at
+                # interval_in_hours hours from start_date.
+                self.instance.last_run_date = self.cleaned_data['start_date']
```
jezdez: This doesn't seem right to me, and I'm not sure if we should allow past start dates at all, and if we allow them we can't set `last_run_date` with it since we rely on that field to know if a job is expired and if it should be run.

The goal of this is to reset the job so as soon as its start date is changed and we literally reset the clock. That means simply setting `last_run_date` to `None` would work for this. That allows us to have accurate information of when a set of job parameters have run last and the user's expectation is met ("If I set a past start date, the job will be run immediately. But If I set a future date, the job will start at that time.")

maurodoglio: Yeah I wasn't sure about that. I think we should prevent the user from setting the start_date to the past, otherwise the user may be confused by the 'run immediately' behavior.

atmo/jobs/forms.py
```diff
@@ -153,12 +154,38 @@ class EditSparkJobForm(BaseSparkJobForm):
...
         )
 
+    def save(self):
+        now = timezone.now()
+        if 'start_date' in self.changed_data:
+            if self.cleaned_data['start_date'] > now:
+                # If the start_date changed and it's set in the future
+                # reset the last_run_date so that it runs at start_date.
+                self.instance.last_run_date = None
+                print 'setting to none'
```
maurodoglio: Ooops

atmo/jobs/forms.py
```diff
@@ -153,12 +154,38 @@ class EditSparkJobForm(BaseSparkJobForm):
...
+                   'Changing this field will reset the job schedule.'),
+    )
+
     def __init__(self, *args, **kwargs):
         super(EditSparkJobForm, self).__init__(*args, **kwargs)
         self.fields['notebook'].help_text += (
             '<br />Current notebook: <strong>%s</strong>' % self.instance.notebook_name
         )
 
+    def save(self):
```
maurodoglio: This feedback is suuuper useful. Thanks!

maurodoglio:
This is ready for another look. The datepicker is initialized in a central place and I thought it's not worth it to specify a different behaviour only for the start_date field of the edit-job form.

tests/test_jobs.py
```diff
@@ -152,7 +160,8 @@ def test_edit_spark_job(request, mocker, client, test_user):
     assert 'form' in response.context
     assert response.context['form'].errors
 
-    edit_data['edit-start_date'] = '2016-03-08 11:17:35'  # fix the date
+    # fix the date
```
jezdez: you can remove the comment here

atmo/jobs/forms.py
```diff
@@ -118,13 +119,14 @@ def clean_notebook(self):
...
 
-    def save(self):
+    def save(self, commit=True):
         # create the model without committing, since we haven't
         # set the required created_by field yet
         spark_job = super(BaseSparkJobForm, self).save(commit=False)
 
-        # actually save the scheduled Spark job, and return the model object
-        spark_job.save(self.cleaned_data['notebook'])
+        if commit:
```
jezdez: Ah, this is where this breaks apart a bit, assuming someone would call this BaseSparkJobForm.save with commit=False, the notebook would never be saved, even if the save would be committed after the fact.

A slightly better version to this I think is to move the call to `scheduling.spark_job_add` from the `SparkJob.save` model save method to the one in the form, so it's clear that passing in `commit=False` will have this effect and not miss the connection.

PR #72, opened 2016-11-09
Bug 1315681 - Use pagination to fetch lists from AWS

jezdez:


jezdez:
@maurodoglio r?

atmo/clusters/jobs.py
```diff
@@ -72,9 +72,11 @@ def update_clusters_info():
     """
     # only update the cluster info for clusters that are pending
     active_clusters = Cluster.objects.active()
+    # get the datetime of the oldest active cluster, set to the start of the day
```
maurodoglio: Why do you need it set to the start of the day instead of the earliest active cluster start_date?

atmo/clusters/jobs.py
```diff
@@ -72,9 +72,11 @@ def update_clusters_info():
...
+    oldest_start_date = active_clusters.datetimes('start_date', 'day')
```
maurodoglio: According to the docs, `datetimes()` returns a queryset, so `oldest_start_date` will evaluate to a list. You may want  to check that the list has at least one item and eventually pass it to `cluster_list()`.

atmo/clusters/jobs.py
jezdez: The start_date in our database is not the same as the date that AWS filters cluster by, we don't sync it. To prevent race conditions in which our start date is a bit earlier due to latency via the API call I think it's safest to just check the days servers.

atmo/clusters/jobs.py
jezdez: Hm, true, will do.

PR #74, opened 2016-11-11
Bug 1315303 - Move RavenJS initialization into own file

jezdez:
Also remove unsafe-inline from CSP script-src setting


jezdez:
Just a PR for records sake. r=me

PR #81, opened 2016-11-22
Bug 1319545 - Add CSP report URI to settings

g-k:
Add a CSP report endpoint that will be handled in nginx.

In dev, this will send to an endpoint that 404s so it could be moved in to stage/prod specific config.


jezdez:
🔒 Nice!

PR #86, opened 2016-11-25
Bug 1311634 - Adopt more dockerflow patterns

jezdez:
- move to using NPM/Yarn for frontend dependencies
- use official Python docker image
- get rid of pipstrap
- use correct user and group IDs

This depends on https://github.com/mozilla-services/cloudops-deployment/pull/356 now

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-analysis-service/86)


jezdez:
@maurodoglio While this took a looooot longer to figure out, this is now ready for review.
@robotblake I need you to take a look at this as well, since this changes the docker setup quite a lot, in the spirit of dockerflow mostly. I'm happy to share extensive debugging experience with you and lots of ideas how we need to update dockerflow to work well with complex set ups like atmo.

bin/run
```diff
@@ -15,17 +15,22 @@ usage() {
 case $1 in
   localweb)
     python manage.py migrate
-    python manage.py runserver 0.0.0.0:${PORT}
+    exec python manage.py runserver 0.0.0.0:${PORT}
```
maurodoglio: Why did you add exec?

docker-compose.override.yml
```diff
@@ -0,0 +1,11 @@
+version: '2'
```
maurodoglio: Is this file used anywhere?

Dockerfile
```diff
@@ -1,40 +1,59 @@
...
-RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        build-essential python python-dev python-pip libpq-dev \
-        postgresql-client gettext sqlite3 libffi-dev \
-    && apt-get autoremove -y \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
+# add a non-privileged user for installing and running the application
+RUN mkdir /app && \
```
maurodoglio: You could let `useradd` create the home dir using the `--create-home` parameter.

bin/run
jezdez: Without it the process doesn't become process ID 1 in the container which would prevent it from correctly receiving posix signals which is required to properly shutdown when the container is stopped or killed.

docker-compose.override.yml
jezdez: Yeah, it's automatically used by docker-compose.yml when not providing explicit docker-compose config files: https://docs.docker.com/compose/extends/#/understanding-multiple-compose-files

Dockerfile
jezdez: Pre-creating the home directory as empty and with required permissions is to avoid `useradd` populating it with home directory skeleton files. Those files aren’t needed and best to leave out so they don’t interfere with anything at a later command.

Dockerfile
```diff
@@ -1,40 +1,59 @@
...
+    chown 10001:0 /app && \
+    chmod g+w /app && \
+    groupadd --gid 10001 app && \
+    useradd --uid 10001 --gid 0 --home /app app
```
jasonthomas: Why are we setting the primary gid for this user to 0?

Dockerfile
```diff
@@ -1,40 +1,59 @@
...
-    && apt-get install -y --no-install-recommends \
-        build-essential python python-dev python-pip libpq-dev \
-        postgresql-client gettext sqlite3 libffi-dev \
-    && apt-get autoremove -y \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
+# add a non-privileged user for installing and running the application
+RUN mkdir /app && \
+    chown 10001:0 /app && \
```
jasonthomas: Does the /app directory need to be owned by the user? Does the app user need to write directly into this directory?

Dockerfile
```diff
@@ -1,40 +1,59 @@
...
-    && pip install -r /tmp/requirements.txt \
-    && find /usr/local -type f -name '*.pyc' -name '*.pyo' -delete \
-    && rm -rf ~/.cache/
-
 COPY . /app
 
-RUN chown webdev.webdev -R .
-USER webdev
+RUN chown -R 10001:0 /app && \
+    chmod -R g+w /app
```
jasonthomas: Why does the group need write access to all files in this directory?

Dockerfile
jezdez: Yeah, I think so, it's implied in [Dockerflow](https://github.com/mozilla-services/Dockerflow) and used that in https://github.com/mozilla-services/Dockerflow/blob/master/Dockerfile#L3-L6 and other Dockerfiles I've seen in the mozilla GitHub org.

That's not to say that It think that's a particular good choice though, since having $HOME and the folder we mount as a volume during development (/app) being the same can lead to all kinds of funny side effects when some command runs that may want to write to $HOME.

What I think we should do (and update Dockerflow to do that), is separate $HOME for the app user from the WORKDIR, e.g. `/opt/src` for the WORKDIR and `/app` as $HOME.

Dockerfile
jezdez: According to a developer for RedHat's OpenShift platform that uses Docker extensively some Docker hosters won't allow us to run a Docker image as root and won't trust our use of USER either since we could map the root user inside the container to whatever user we want and then the hoster is screwed. So they ignore USER and assign a user ID that is different from other projects hosted there to the running container. But since that user ID isn't matching a real UNIX account it ends up setting the group ID to 0.

In other words, if we chose to use a group ID of 0, we're making sure our image is as compatible as possible to various Docker hosting services. If we don't strive to make Dockerflow host agnostic then we can update this to create an own group for the app user.

Dockerfile
jezdez: That's again a result of trying to make the image compatible to restrictive hosters. If we run it in an environment that uses the USER command, then it can write by way of being owner.

But when the uid is overridden by the hoster to a random and high uid, we can rely on the fact that processes run as group root and that directories and files are writable as the root group.

jezdez: Same applies as above, if we decide that we don't want to make Dockerflow agnostic to the hoster and introduce an own group for the app user we can probably remove that again. Mind you that this can again have an interesting overlap with the duplicity of /app being $HOME and target for a mounted volume during development, since volumes are currently not mounted with the USER permissions but with root.

Dockerfile
jasonthomas: I can't comment on other docker hosting provider requirements (I haven't worked much with them) but we should aim to make our Dockerfile compatible for our use cases first. If in the future we decide to go another route we discuss it then. Let's use group id other 0.

Dockerfile
jasonthomas: This Dockerfile should represent the Docker image used for production deployments. We should use a different Dockerfile for development purposes.

Dockerfile
jasonthomas: As discussed we should use a different Dockerfile for development purposes.

jezdez:
Phew, getting the coverage reporting working was a bit of a mess, but this is ready for another pass. @jasonthomas @maurodoglio r?

Dockerfile
```diff
@@ -1,40 +1,57 @@
...
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Install node from NodeSource
+RUN curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
+    echo 'deb https://deb.nodesource.com/node_4.x jessie main' > /etc/apt/sources.list.d/nodesource.list && \
+    echo 'deb-src https://deb.nodesource.com/node_4.x jessie main' >> /etc/apt/sources.list.d/nodesource.list && \
+    apt-get update && apt-get install -y nodejs
+
+# Create static and npm roots
+RUN mkdir -p /opt/npm /opt/static && \
```
jasonthomas: What is /opt/static used for?

Dockerfile
```diff
@@ -1,40 +1,57 @@
...
+WORKDIR /tmp
+RUN pip install --upgrade --no-cache-dir -r requirements.txt
+
+# Install frontend dependencies using NPM
+COPY package.json /opt/npm/
+
+# Switch to /opt/npm to install dependencies outside home dir
+WORKDIR /opt/npm
+RUN npm install && \
+    chown -R 10001:10001 /opt/npm && \
```
jasonthomas: Does '10001' user need to be able to write to the files in /opt/npm?

Dockerfile
```diff
@@ -1,40 +1,57 @@
...
-    && pip install -U pip setuptools \
-    && pip install -r /tmp/requirements.txt \
-    && find /usr/local -type f -name '*.pyc' -name '*.pyo' -delete \
-    && rm -rf ~/.cache/
-
 COPY . /app
 
-RUN chown webdev.webdev -R .
-USER webdev
+RUN chown -R 10001:10001 /app
```
jasonthomas: Other than /app/static and /app/media (for collectstatic management command), why else would user 10001 need write access to this path?

Dockerfile
```diff
@@ -1,40 +1,57 @@
...
+
+# Create static and npm roots
+RUN mkdir -p /opt/npm /opt/static && \
+    chown -R 10001:10001 /opt
+
+# Install Python dependencies
+COPY requirements.txt /tmp/
+# Switch to /tmp to install dependencies outside home dir
+WORKDIR /tmp
+RUN pip install --upgrade --no-cache-dir -r requirements.txt
```
jasonthomas: Does the --upgrade flag here do anything since with are pinning versions in requirements.txt?

Dockerfile
jezdez: That's the location where collectstatic is being run both when the tests are run and [during deployment](https://github.com/mozilla-services/cloudops-deployment/blob/ecadfaf825f127c3fee3d82e780c782392c63037/projects/atmo/puppet/modules/atmo/manifests/app.pp#L22-L30).

Dockerfile
jezdez: Nope, good catch, this is basically a leftover from when we didn't use pinned versions.. removing it.

Dockerfile
jezdez: Hm, no only reading. During the run of Django's collectstatic management command it'll try to find the `node_modules` in that folder and copy it to the right place using [django-npm](https://github.com/kevin1024/django-npm).

Dockerfile
jezdez: Hm, it doesn't need write access, not even for collectstatic since we're using /opt/static for that, *not* /app/static as that collides with the volume mounting during development. And I'd rather use the same location for dev and prod (/opt/static).

jezdez:
This depends on https://github.com/mozilla-services/cloudops-deployment/pull/356 now

Dockerfile
jasonthomas: Okay makes sense.

Dockerfile
```diff
@@ -1,40 +1,66 @@
...
 
 COPY . /app
 
-RUN chown webdev.webdev -R .
-USER webdev
+RUN chown -R 10001:10001 /app
+
+USER 10001
+
+RUN python manage.py collectstatic --noinput
```
jasonthomas: I tried to build the container and I received the following error:

```
Step 20 : RUN python manage.py collectstatic --noinput
 ---> Running in d19bbee75d22
Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 185, in fetch_command
    settings.INSTALLED_APPS
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 55, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 43, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 120, in __init__
    raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

I moved .env-dist to .env and added 'DJANGO_SECRET_KEY' to some random value but it looks like it is being blocked on connections to redis:

```
strace -fp 22522
strace: Process 22522 attached
connect(3, {sa_family=AF_INET, sin_port=htons(6379), sin_addr=inet_addr("104.239.213.7")}, 16) = -1 ETIMEDOUT (Connection timed out)
```

Any reason why collectstatic management command needs access to redis?

Dockerfile
jasonthomas: Currently files in the /app directory have 664 permissions and directories have 775. I think we need to restrict these permissions, such that the app user has no write capability via user or group. If we remove the `chown` command the files and directories will be owned by root:root. We could also add a chmod -R g-w /app to remove group write permissions too.

Dockerfile
jezdez: @jasonthomas Ah, good point, the settings were not prepared to run without the rest of the compose system be up. I just fixed that by adding a build-only configuration, which will be only used for the collectstatic call.

The Redis issue is also related to that -- our code expected the Redis connection to success on startup (including collectstatic) since we've always run the tests inside of the compose environment (still do). It just happens to have bubbled up because the collectstatic call in the Dockerfile is outside that environment. I've fixed that now by wrapping the Redis call during startup with connection error handling.

jezdez:
@jasonthomas Any chance to take a look at this again?

jasonthomas:
I tried building this again but ran into a issue with npm dependency resolution:

```
Step 14 : RUN npm install &&     chown -R 10001:10001 /opt/npm &&     npm cache clean
 ---> Running in 19316f2e00f0
npm WARN peerDependencies The peer dependency jquery@^1.8.3 || ^2.0 included from eonasdan-bootstrap-datetimepicker will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm WARN peerDependencies The peer dependency moment-timezone@^0.4.0 included from eonasdan-bootstrap-datetimepicker will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.
npm ERR! Linux 4.3.5-300.fc23.x86_64
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "install"
npm ERR! node v4.7.2
npm ERR! npm  v2.15.11
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package jquery@3.1.1 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer eonasdan-bootstrap-datetimepicker@4.17.45 wants jquery@^1.8.3 || ^2.0

npm ERR! Please include the following file with any support request:
npm ERR!     /opt/npm/npm-debug.log
```

jezdez:
@jasonthomas Meh, this was an issue introduced in https://github.com/Eonasdan/bootstrap-datetimepicker/commit/86b24d33a9b8b0b9473f95d9c54cb4322f4a1fc3

jezdez:
@jasonthomas Sorry for the mess, this is now fixed. Mind reviewing again?

jasonthomas:
The npm looks fixed \o/

`Step 19 : RUN DJANGO_CONFIGURATION=Build &&     python manage.py collectstatic --noinput` looks like it still hangs on redis:

`connect(3, {sa_family=AF_INET, sin_port=htons(6379), sin_addr=inet_addr("104.239.213.7")}, 16) = -1 ETIMEDOUT (Connection timed out)`

jezdez:
@jasonthomas I can't reproduce this error, the build works for me, and I did:

- stop all docker containers
- force removed all images
- ran `docker build -t app:build .` like circle ci does

I ran this both on a Mac with Docker.app as well as a Ubuntu 16.04 machine running on Digital Ocean.

Can you share the full traceback of your build?
I noticed the IP 104.239.213.7 in your error, which seems to be one from Rackspace. Can you elaborate where that comes from?
Which versions of Docker/docker-compose are you running?

jasonthomas:
Looks like cleaning up my old images worked. r+

PR #121, opened 2017-01-14
Bug 1322759 - Add description field to spark jobs

robhudson:
The new field as it appears on the form...

screenshot 2017-01-13 18 12 11


atmo/forms/mixins.py
```diff
@@ -1,10 +1,10 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, you can obtain one at http://mozilla.org/MPL/2.0/.
-from collections import OrderedDict
 import uuid
+from collections import OrderedDict
+
 from django import forms
-from django.core.exceptions import ImproperlyConfigured
```
jezdez: ❤️

atmo/jobs/forms.py
```diff
@@ -26,9 +26,20 @@ class BaseSparkJobForm(FormControlFormMixin, CachedFileModelFormMixin,
             'class': 'identifier-taken-check',
             'data-identifier-taken-check-url': reverse_lazy('jobs-identifier-taken'),
         }),
-        help_text='A brief description of the scheduled Spark job\'s purpose, '
+        help_text='A slug-like string to identify your Spark job, '
```
jezdez: "slug-like" is a bit technical and not likely to be understood by the ATMO users, can we find a middleground here?

atmo/jobs/models.py
```diff
@@ -38,6 +38,10 @@ class SparkJob(EMRReleaseModel, CreatedByModel):
         help_text="Job name, used to uniqely identify individual jobs.",
         unique=True,
     )
+    description = models.TextField(
+        help_text='Job description.',
+        default='',
```
jezdez: Is the empty default needed for this?

atmo/jobs/forms.py
robhudson: Question: Does this string need to be a slug? Or can it just be a short string with spaces or capitals? I didn't see any slug processing but all the examples in tests are "slug-like".

atmo/jobs/models.py
robhudson: That default is needed for the migration for existing records in the database. Is there a better place for that to live?

atmo/jobs/forms.py
jezdez: Yeah, the [job form will validate](https://github.com/mozilla/telemetry-analysis-service/blob/7ce65254adf96217931ca3f90d0b6ef9970b9e73/atmo/jobs/forms.py#L23) that with `[\w-]{1,100}$` and [remove whitespaces and underscores on the client side](https://github.com/mozilla/telemetry-analysis-service/blob/7ce65254adf96217931ca3f90d0b6ef9970b9e73/atmo/static/js/forms.js#L36-L37) when it checks for identifier availability.

So I'd say it *is* slug-like, but think the term is not easy to understand -- hence the more general term "identifier". Maybe we should call it "label-like"?

atmo/jobs/models.py
jezdez: Ah, nope, totally blanked on that one. Carry one :)

robhudson:
The help text has been updated. Let me know if that works.

PR #173, opened 2017-01-31
Bug 1328892 - Refactor spark job scheduling to use spot instances.

jezdez:
Bug 1328892 - Refactor spark job scheduling to use spot instances.

This does:

- introduces Provisioner classes for clusters and Spark jobs, that share some common code but extend the calls to EMR according to their task
- extends the tests for jobs and cluster code, fixing issues with the code where found
- prepare the Spark job code to refer to a 1:1 relationship with a cluster and its provisioner if needed


jezdez:
There is lot of moving around, I'm going to cut the refactor into smaller commit next

atmo/provisioners.py
```diff
@@ -0,0 +1,108 @@
...
+            'ReleaseLabel': 'emr-%s' % emr_release,
+            'Configurations': self.spark_emr_configuration(),
+            'Instances': {
+                'InstanceGroups': instance_groups,
+                'Ec2KeyName': self.config['EC2_KEY_NAME'],
+                'KeepJobFlowAliveWhenNoSteps': True,  # same as no-auto-terminate
+            },
+            # TODO: this was missing in the job scheduling, but existing in the
+            # cluster provisioning paramters, should this be left out of jobs?
+            'JobFlowRole': self.config['SPARK_INSTANCE_PROFILE'],
```
jezdez: @maurodoglio This paramater to the `run_job_flow` EMR endpoint is missing in the `scheduling.py` file which deals with running the scheduled spark jobs. Do you happen to know what its purpose is bigger than what in the Boto docs is?

maurodoglio: JobFlowRole? I see it's passed to run_job_flow in both spark_job_run and cluster_start on master. My understanding is that it defines the IAM role (aka aws permissions) for each node in the cluster.

jezdez: Uuh, nevermind, you're totally right. I must have missed this during the refactor, thanks for the clarification :)

atmo/clusters/provisioners.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ClusterProvisioner(Provisioner):
+    log_dir = 'clusters'
+
+    def start(self, user_email, identifier, size, emr_release,
+              public_key):
+        """
+        Given a user's email, a cluster identifier, a worker count, and a user
+        public key, spawns a cluster with the desired properties and returns
+        the jobflow ID.
```
robhudson: The comment doesn't match the args but I think it might be better if they didn't anyway. Maybe remove them in the comment? Easier to maintain.

robhudson:
I looked over this and like what I see. I'm not very familiar with this part of the codebase so someone who is may have more comments.

atmo/clusters/provisioners.py
jezdez: Good idea! Thanks :)

atmo/keys/models.py
```diff
@@ -56,6 +56,5 @@ def prefix(self):
         return self.key.strip().split()[0]
 
     def save(self, *args, **kwargs):
-        if not self.fingerprint and self.key:
-            self.fingerprint = calculate_fingerprint(self.key)
+        self.fingerprint = calculate_fingerprint(self.key)
```
maurodoglio: Is this related to this commit?

tests/blockade.py
```diff
@@ -0,0 +1,75 @@
+"""
+This blocks all HTTP requests via httplib (which includes requests).
+
+this is a fixed copy of pytest-blockage, called "blockade"
```
jezdez: Yeah, there is already a half-assed version called [pytest-blockage](https://pypi.python.org/pypi/pytest-blockage) that this is based on, I think it would make sense to release this eventually

pytest.ini
```diff
@@ -3,3 +3,5 @@ norecursedirs = .git .* static
 addopts = -rsxX --showlocals --tb=native --nomigrations
 DJANGO_SETTINGS_MODULE = atmo.settings
 DJANGO_CONFIGURATION = Test
+blockade = True
+; blockade-http-whitelist = accounts.google.com
```
jezdez: Yeah, mostly a leftover, but it's also a reminder that this is possible. We need some proper docs eventually

atmo/clusters/models.py
```diff
@@ -116,6 +116,31 @@ def __str__(self):
     def __repr__(self):
         return "<Cluster {} of size {}>".format(self.identifier, self.size)
 
+    @property
```
jezdez: Yeah, this is about readability, I always considered properties to belong right after the magic methods, will fix the commit message accordingly

atmo/settings.py
```diff
@@ -428,6 +428,11 @@ def post_setup(cls):
 
 class Test(Dev):
     """Configuration to be used during testing"""
+    CONSTANCE_BACKEND = 'constance.backends.redisd.RedisBackend'
```
jezdez: So far we've treated Redis as a non-persistent data store and the database as the single source of truth, which was the reason to use the database backend (with the Redis cache as the cache in front). I'm not sure if we should use the Redis backend for constance simply because I don't know how safe the AWS Redis cluster we use is.

tests/conftest.py
```diff
@@ -8,9 +8,14 @@
 from django.utils import timezone
 from django.core.files.uploadedfile import InMemoryUploadedFile
 
+from atmo.clusters.provisioners import ClusterProvisioner
+from atmo.jobs.provisioners import SparkJobProvisioner
 from atmo.keys.models import SSHKey
 
 
+pytest_plugins = ['blockade']
```
jezdez: Ah, yeah, I can do that. Good point.

atmo/keys/models.py
jezdez: Yeah, the commit tests the behavior, which we change here of always updating the fingerprint based on the key value, https://github.com/mozilla/telemetry-analysis-service/pull/173/files#diff-da6b35707e8a7f55bbf2ba043124f462R61

The reasoning here is that we don't want to accidentally update the key somehow and not be able to verify the fingerprint separately. In other words, there is no reason to have the fingerprint calculation happen conditionally and only on the first save.

jezdez: Should I separate this change into an own commit?

maurodoglio: Yes please 😺 . And could you also copy&paste the second part of [your comment above](https://github.com/mozilla/telemetry-analysis-service/pull/173#discussion_r99821763) into the commit message? It would be great for posterity

jezdez: Will do, good idea!

PR #182, opened 2017-02-09
Bug 1328892 - Fix regression with KeepJobFlowAliveWhenNoSteps that prevented scheduled Spark job clusters to not auto-terminate after all steps had been run.

jezdez:


tests/test_provisioners.py
```diff
@@ -424,7 +425,7 @@ def test_spark_job_run(mocker, is_public, spark_job_provisioner):
                     'Name': 'Master',
                 }
             ],
-            'KeepJobFlowAliveWhenNoSteps': True,
+            'KeepJobFlowAliveWhenNoSteps': False,
```
robhudson: I would expect a similar check for cluster creation.

atmo/jobs/provisioners.py
```diff
@@ -19,6 +19,12 @@ def __init__(self):
         # the S3 URI to the job shell script
         self.batch_uri = 's3://%s/steps/batch.sh' % self.config['SPARK_EMR_BUCKET']
 
+    def job_flow_params(self, *args, **kwargs):
+        params = super(SparkJobProvisioner, self).job_flow_params(*args, **kwargs)
+        # auto-terminate the job cluster after all steps have been done
+        params.setdefault('Instances', {})['KeepJobFlowAliveWhenNoSteps'] = False
+        return params
+
```
robhudson: If you made `False` the default in the base Provisioner class you could skip this. Being `False` by default is probably a good safe-guard. But if you prefer this that's fine too. What does AWS do in absence of this?

tests/test_provisioners.py
jezdez: Added that now

PR #185, opened 2017-02-10
Issue #189 - Store history of Spark job runs

robhudson:


atmo/jobs/models.py
```diff
@@ -120,13 +112,15 @@ def has_never_run(self):
         Looks at both the cluster status and our own record when
         we asked it to run.
         """
-        return (self.most_recent_status == self.DEFAULT_STATUS or
-                self.last_run_date is None)
+        return (self.latest_run is None or
+                self.latest_run.status == DEFAULT_STATUS or
+                self.latest_run.run_date is None)
```
jezdez: This may be easier to say `self.runs.exclude(status=DEFAULT_STATUS).exists()`? I now wonder if we should have a spark job run *always* have a non-empty status. Even if it's "CREATED" only, so we can filter them out.

atmo/jobs/models.py
```diff
@@ -171,8 +176,9 @@ def update_status(self):
         """
         info = self.get_info()
         if info is not None:
-            self.most_recent_status = info['state']
-        return self.most_recent_status
+            (SparkJobHistory.objects.filter(jobflow_id=self.last_jobflow_id)
+                            .update(most_recent_status=info['state']))
+        return info['state']
```
jezdez: Let's more the update_status method into the run model, since that's where we really store it

atmo/jobs/models.py
```diff
@@ -204,14 +211,18 @@ def run(self):
             is_public=self.is_public,
             job_timeout=self.job_timeout,
         )
-        self.last_run_date = timezone.now()
+        # Create new job history record.
+        SparkJobHistory.objects.create(
```
jezdez: Let's use the reverse relationship here, `self.runs.create(...)`

atmo/jobs/models.py
```diff
@@ -76,21 +78,11 @@ class SparkJob(EMRReleaseModel, CreatedByModel):
         default=True,
         help_text="Whether the job should run or not."
     )
-    last_run_date = models.DateTimeField(
-        blank=True,
-        null=True,
-        help_text="Date/time that the job was last started, null if never."
-    )
-    current_run_jobflow_id = models.CharField(
+    last_jobflow_id = models.CharField(
```
robhudson: Right. I started w/o it but there was one case where it was nice to not have to do a query to get it.

atmo/jobs/models.py
```diff
@@ -226,3 +237,32 @@ def delete(self, *args, **kwargs):
 
     def get_results(self):
         return self.provisioner.results(self.identifier, self.is_public)
+
+
+@python_2_unicode_compatible
+class SparkJobHistory(EditedAtModel):
```
robhudson: 👍 I like that

atmo/jobs/models.py
```diff
@@ -226,3 +237,32 @@ def delete(self, *args, **kwargs):
...
+        max_length=50,
+        blank=True,
+        null=True,
+    )
+    run_date = models.DateTimeField(
+        blank=True,
+        null=True,
+        help_text="Date/time that the job was started.",
+    )
+    most_recent_status = models.CharField(
```
robhudson: I started with that too. But since this gets updated whenever the rq job runs and updates all the statuses of the jobs, it will change as the lifecycle of the cluster changes. So it really is the most recent status of this particular job run.

atmo/jobs/models.py
robhudson: Do we not want to update the status each time the rq `run_jobs` runs?

atmo/jobs/models.py
```diff
@@ -204,14 +211,18 @@ def run(self):
             is_public=self.is_public,
             job_timeout=self.job_timeout,
         )
-        self.last_run_date = timezone.now()
+        # Create new job history record.
+        SparkJobHistory.objects.create(
+            spark_job=self,
+            jobflow_id=jobflow_id,
+            run_date = timezone.now(),
```
robhudson: How else could this be handled? Right now it's not much different than what is in `created_at`.

atmo/jobs/models.py
robhudson: Oh nvm, I understand what you're saying now.

atmo/jobs/forms.py
```diff
@@ -221,10 +221,13 @@ def clean_start_date(self):
     def save(self, commit=True):
         obj = super(EditSparkJobForm, self).save(commit=False)
         if 'start_date' in self.changed_data:
```
robhudson: @jezdez See below. What should I do here?

atmo/jobs/models.py
jezdez: Right, let's keep it that way and then reconsider as part of #205

atmo/jobs/models.py
jezdez: Right, good point 👍

atmo/jobs/forms.py
jezdez: @robhudson Ah, I just replied to your comments, did you mean that with "below"?

robhudson: No sorry, I mean with the "TODO" in this file. On `save` currently we reset the `last_run_date` if the `start_date` is edited in the form. With historical job runs this doesn't make sense to me and I would think this should be removed. But I wanted to double check.

atmo/jobs/models.py
robhudson: I tried the `.exists()` query but since `latest_run` is cached on the model doing these few checks is easier for testing vs having to save to the db. I think it's also more clear than a query and matches the intention of the method better.

I could easily make `DEFAULT_STATUS = 'CREATED'` if you'd like.

atmo/jobs/forms.py
robhudson: @jezdez Can I bother you for an answer hear today? ^

jezdez: @robhudon Totally, apologies, you're right, we can fetch that info in the future with `spark_job.runs.datetimes('run_date', 'second', order='DESC')[0]` or `spark_job.runs.latest().run_date`.

atmo/jobs/models.py
```diff
@@ -226,3 +213,53 @@ def delete(self, *args, **kwargs):
 
     def get_results(self):
         return self.provisioner.results(self.identifier, self.is_public)
+
+
+@python_2_unicode_compatible
```
jezdez: We don't really need this anymore, I think. Since we're Python 3 only now.

atmo/jobs/models.py
```diff
@@ -204,14 +187,18 @@ def run(self):
...
-        self.last_run_date = timezone.now()
-        self.update_status()
-        self.save()
+        # Create new job history record.
+        run = self.runs.create(
+            spark_job=self,
+            jobflow_id=jobflow_id,
+            scheduled_date = timezone.now(),
+        )
+        run.update_status()
```
jezdez: Once the job run has been created we need to invalidate the cached_property of `self.latest_run` since it could otherwise lead to race conditions

jezdez: You can do that by calling `delattr(self, 'latest_run')` or `del self.latest_run`. It may raise an AttributeError though if it wasn't in fact populated before, so wrap it in a try: except block.

tests/test_jobs.py
```diff
@@ -146,13 +145,15 @@ def test_create_spark_job(client, mocker, notebook_maker,
         size=spark_job.size,
         user_email=test_user.email,
     )
+    spark_job.latest_run = spark_job.get_latest_run()
```
jezdez: Why do you set the `latest_run` attribute here with the result of `get_latest_run`?

atmo/jobs/jobs.py
```diff
@@ -13,9 +13,9 @@ def run_jobs():
     """
     run_jobs = []
     for job in SparkJob.objects.all():
-        # first let's update the status to update most_recent_status
-        job.update_status()
-        job.save()
+        # first let's update the status if there is a prior run
+        if job.latest_run:
```
jezdez: Going through all jobs and fetching all latest runs can be expensive the more jobs and runs we add. Can we find a different way to iterate over the job runs here?

atmo/jobs/migrations/0012_add_job_run_history.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+
+
+def rollback_job_runs(apps, schema_editor):
+    SparkJob = apps.get_model('jobs', 'SparkJob')
+    SparkJobRun = apps.get_model('jobs', 'SparkJobRun')
+
+    for job in SparkJob.objects.all():
+        last_run = SparkJobRun.objects.filter(spark_job=job).latest()
+        if last_run:
```
jezdez: I think this may raise a `DoesNotExist` exception if there is for some reason no latest job run for the job, e.g. because the current_run_jobflow_id was not set and in the forward migration no job run was created.

atmo/jobs/models.py
```diff
@@ -226,3 +213,53 @@ def delete(self, *args, **kwargs):
...
     def get_results(self):
         return self.provisioner.results(self.identifier, self.is_public)
+
+
+@python_2_unicode_compatible
+class SparkJobRun(EditedAtModel):
+
+    spark_job = models.ForeignKey(
+        SparkJob,
+        on_delete=models.CASCADE,
```
jezdez: So I'm wondering if we should not have this cascade but set to NULL instead to be able to keep records of job runs, even if the users delete the jobs in the frontend. Do you think we can find a pattern to clean up differently?

atmo/templates/atmo/jobs/detail.html
```diff
@@ -164,7 +164,7 @@
       <dd>{{ spark_job.end_date }}</dd>
       {% endif %}
       <dt>Last run date</dt>
```
jezdez: Mind s/run date/scheduled date/g here?

atmo/jobs/models.py
robhudson: Yeah but they are on every model class. I thought a separate commit could take care of it.

tests/test_jobs.py
robhudson: Because `spark_job.run()` creates a new latest run that needed to be forcefully updated on the `spark_job` object. This will be fixed by doing what you say to do above, which is to invalidate the old `latest_run`. I like that better.

atmo/jobs/jobs.py
robhudson: I can't think of anything and being a simple query won't it be pretty quick? If we put an index on the date column that would help once there are lots of runs.

Otherwise perhaps go back to adding the denormalized `last_run_id` column on `SparkJob` so we can query them with a JOIN in a single query?

atmo/jobs/migrations/0012_add_job_run_history.py
robhudson: You're right. I've added a `try` block around it. Thanks.

atmo/jobs/models.py
robhudson: I'm curious why we might need this? Would keeping a history be useful for calculations?

One idea would be to have a `deleted` flag on the `SparkJob` and update our queries to `SELECT ... WHERE deleted=false` and not actually delete the record but stop showing it to the user if it were deleted. That way we have all the relevant data in the system to perform calculations.

atmo/templates/atmo/jobs/detail.html
robhudson: Yep, and I added the new date fields here also.

robhudson:
Updated based on comments. Added extra datetime fields per #205 with tests.

atmo/jobs/models.py
jezdez: Yeah, basically we wouldn't want to lose the data, even while we may want to allow users to delete the job. We currently don't delete job notebook files from S3 either, so there is already a cleanup required to be done eventually. In terms of history tracking, I'd rather have a bit too much info (especially since it doesn't contain any private info) that too little. Then once we built a graph in Redash against the table we can decide how far back we want to keep the data for sure, then then write the cleanup script if needed.

atmo/jobs/models.py
jezdez: Fair enough, let's do this separately. I've filed #244 for this.

PR #228, opened 2017-02-14
Fix #227 - Update `terminate` to correctly terminate cluster

### LINKED ISSUE #227: AttributeError: 'SparkJobProvisioner' object has no attribute 'stop'
While testing locally in a branch I bumped into this. I didn't test it on master but looking at the code this seems like it may also be a problem.

The `SparkJob.terminate` calls `self.provisioner.stop(self.current_run_jobflow_id)`, where `self.provisioner` is a `SparkJobProvisioner`. But currently the `SparkJobProvisioner` has no `stop` method and neither does the parent class, it is only defined on the `ClusterProvisioner`. In my branch I changed the `terminate` method to call `self.cluster_provisioner.stop` instead and it fixed my issue locally.

robhudson:


autolander:

jezdez:
Yikes, that's a good catch! The issue was wrongly introduced in https://github.com/mozilla/telemetry-analysis-service/commit/a7a4c86a85083c51f530c8a2564f6514585dfb9b#diff-8d0f4fb6729f2067a211fb64aa9e32bcR211

PR #246, opened 2017-02-20
Close database connections on startup of the worker, before and after each job.

### LINKED ISSUE #245: Handle dropped database connections gracefully
In https://sentry.prod.mozaws.net/operations/atmo-stage/issues/383510/ we see that the database connection is closed prematurely and neither Django nor psycopg is able to recover from that.

The error is well known to be related to using process fork with open connections: http://pythonhosted.org/psycopg2/usage.html#thread-safety

This is related to https://github.com/ui/django-rq/issues/216 and https://github.com/celery/celery/blob/4d63867c8281e94c74dcdf84fe2a441eaed6671b/celery/fixups/django.py#L135-L192.

jezdez:
Fix #245.

### DIFF
--- a/atmo/worker.py
+++ b/atmo/worker.py
@@ -1,0 +1,26 @@
+from django.db import connections, DatabaseError, InterfaceError
+from rq_retry import RetryWorker
+
+
+class ConnectionClosingRetryWorker(RetryWorker):
+    def close_database(self):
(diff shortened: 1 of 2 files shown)


jezdez:
@maurodoglio There already is a PR open by @depaolim: https://github.com/ui/django-rq/pull/220

PR #258, opened 2017-02-23
Auto-select SSH key if only key

robhudson:


atmo/clusters/views.py
```diff
@@ -25,10 +31,10 @@ def new_cluster(request):
             )
         )
         return redirect('keys-new')
-    initial = {
-        'identifier': '{}-telemetry-analysis'.format(user_display(request.user)),
-        'size': 1,
-    }
+    elif ssh_key_count == 1:
+        # If only 1 ssh key, make it pre-selected.
+        initial['ssh_key'] = request.user.created_sshkeys.values('pk')[0]['pk']
```
jezdez: So I'm a fan of using the new queryset APIs, e.g. `first()` since it returns None if there is none. That way you can combine the multiple calls from line 22 to here.

E.g.:

```python
first_ssh_key = request.user.created_sshkeys.first()  # order by date?

if first_ssh_key is None:
    # message and redirect to new key view
else:
    initial['ssh_key'] = first_ssh_key.pk
```

That'll even work if there are multiple keys available and select the first one. In case you're doing ordering on the base query set, make sure to update the queryset of the `ModelChoiceField` in `cluster/forms.py`

bin/run
```diff
@@ -49,7 +49,7 @@ case $1 in
     python manage.py collectstatic --noinput
     coverage erase
     coverage run -m py.test
-    coverage report
+    coverage report -m
```
jezdez: 👍

tests/test_clusters.py
```diff
@@ -32,6 +34,26 @@ def cluster_provisioner_mocks(mocker):
...
+    assert form.initial['size'] == 1
+    assert form.initial['ssh_key'] == ssh_key.id
+
+
+def test_no_keys_redirect(client, test_user):
+    response = client.post(reverse('clusters-new'), {}, follow=True)
+    assert response.status_code == 200
+    assert response.redirect_chain[-1] == (reverse('keys-new'), 302)
+    assert ('No SSH keys associated to you' in
+            [m for m in get_messages(response.wsgi_request)][0].message)
```
jezdez: I think you can just use `response.request` here. I wonder if pytest-django has a helper for messages?

atmo/clusters/views.py
robhudson: If you're ok with defaulting to an arbitrary ssh key I could do that. I was using `values` so the SQL is more "direct" and to avoid creating a `SSHKey` class just to pluck out the `pk`.

tests/test_clusters.py
robhudson: I'll test using `request.response` instead. I didn't see a helper listed in the pytest django docs. I agree it's kind of ugly.

robhudson: Nope, this expects the format of `wsgi_request`.

atmo/clusters/views.py
robhudson: Optionally we could add a `default` flag? But that seems a little overkill. How many people even have more than one key?

jezdez: @robhudson Yeah, I wouldn't be concerned about speed of the cluster form view given the amount of traffic but would lean towards improving readability. But yeah, it probably doesn't matter. r+

robhudson: Another option that's a simple single line change from what's there now is: `request.user.created_sshkeys.get().pk` since we know there's only 1 already.

PR #288, opened 2017-03-07
Remove EMR release 4.5.0

vitillo:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1344020. Note that we currently have one scheduled job left running on 4.5.0 owned by @Dexterp37 which has to be ported to a newer EMR release.


Dexterp37:
@vitillo , how much time do I have to port that?

vitillo:
@Dexterp37 ideally asap as without it we can't land the centralized metastore. It should be rather painless though and all you might have to do is simply change EMR release. Please check the [Spark 2 release notes](https://spark.apache.org/releases/spark-release-2-0-0.html) for breaking changes.

mreid-moz:
In the future, it might be a good idea to mark old releases as deprecated for some time before removing them.

jezdez:
@mreid-moz Hm, marking as deprecated, e.g. by having the label read "4.5.0 (deprecated)" in the UI? I think that'd only work if people get notified upon deprecation since they rarely visit the job editing page again.

Dexterp37:
The blocker is being taken care of in mozilla/mozilla-reports#36

jezdez:
@Dexterp37 Yay!

jezdez:
@Dexterp37 To be clear, https://analysis.telemetry.mozilla.org/jobs/2/ still exists, just making sure we're good that deployment can go ahead.

Dexterp37:
Thanks for following up @jezdez ! No, unfortunately we are not good to deploy until the other PR is merged and the job you referenced gets killed. Otherwise, we risk of loosing churn data that's being used :(

jezdez:
Okay, makes sense. I'm going to approve the changes here though based on that. No data migration needed.

jezdez:
@vitillo Let's wait with merging this until we have the blockers fixed.

mreid-moz:
> Hm, marking as deprecated, e.g. by having the label read "4.5.0 (deprecated)" in the UI? I think that'd only work if people get notified upon deprecation since they rarely visit the job editing page again.

That's what I was thinking, but you're right, it probably wouldn't be effective as a way to bring it to people's attention. Adding the label combined with an announcement or a "what's new" page in ATMO could work. Anyways, it's unrelated to this PR, just figured I'd mention it for future removals.

Dexterp37:
@jezdez @vitillo I've removed the offending job, you're good to go!

jezdez:
@vitillo Just FTR, the email alerts pull request had a conflicting migration (#291) so I fixed this in https://github.com/mozilla/telemetry-analysis-service/commit/13cc232761385c585405a921c818ec8ce79a5f22. I know this was your first Django migration, so this is just a thing to look out when doing multiple features at the same time that change database tables.

vitillo:
@jezdez thanks!

PR #289, opened 2017-03-07
Use ListCluster EMR API to update last job runs statuses.

### LINKED ISSUE #284: Use list_cluster EMR API for updating job run status
We currently facing lots of AWS API throttling when updating individual job run statuses before deciding to run them or not. We basically run one `describe_cluster` call per jobs per call against the `run_jobs` job.

jezdez:
This should fix a lot of the throttling issues since we do less API queries. Fix #284.

### DIFF
--- a/atmo/jobs/jobs.py
+++ b/atmo/jobs/jobs.py
@@ -6,6 +6,7 @@ import logging
 import newrelic.agent
 
 from .models import SparkJob
+from atmo.clusters.models import Cluster
 from atmo.clusters.provisioners import ClusterProvisioner
 
 logger = logging.getLogger(__name__)
@@ -21,17 +22,20 @@ def run_jobs():
     jobs = SparkJob.objects.all()
 
     # get the jobs with prior runs
-    jobs_with_runs = jobs.filter(runs__isnull=False).prefetch_related('runs')
-    logger.debug('Updating Spark jobs: %s', jobs_with_runs)
+    jobs_with_active_runs = jobs.filter(
+        runs__isnull=False,
+        runs__status__in=Cluster.ACTIVE_STATUS_LIST,
+    ).prefetch_related('runs')
+    logger.debug('Updating Spark jobs: %s', jobs_with_active_runs)
 
     # create a map between the jobflow ids of the latest runs and the jobs
     jobflow_job_map = {
         job.latest_run.jobflow_id: job
-        for job in jobs_with_runs
+        for job in jobs_with_active_runs
     }
     # get the created dates of the job runs to limit the ListCluster API call
     provisioner = ClusterProvisioner()
-    runs_created_at = jobs_with_runs.datetimes('runs__created_at', 'day')
+    runs_created_at = jobs_with_active_runs.datetimes('runs__created_at', 'day')
     logger.debug('Fetching clusters older than %s', runs_created_at[0])
 
     cluster_list = provisioner.list(created_after=runs_created_at[0])
--- a/atmo/jobs/jobs.py
+++ b/atmo/jobs/jobs.py
@@ -20,31 +20,31 @@ def run_jobs():
     run_jobs = []
     jobs = SparkJob.objects.all()
 
-    with transaction.atomic():
-        # get the jobs with prior runs
-        jobs_with_runs = jobs.filter(runs__isnull=False).prefetch_related('runs')
-        logger.debug('Updating Spark jobs: %s', jobs_with_runs)
+    # get the jobs with prior runs
+    jobs_with_runs = jobs.filter(runs__isnull=False).prefetch_related('runs')
+    logger.debug('Updating Spark jobs: %s', jobs_with_runs)
 
-        # create a map between the jobflow ids of the latest runs and the jobs
-        jobflow_job_map = {
-            job.latest_run.jobflow_id: job
-            for job in jobs_with_runs
-        }
-        # get the created dates of the job runs to limit the ListCluster API call
-        provisioner = ClusterProvisioner()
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
   worker:
     extends:
       service: app
+    depends_on:
+      - app
```
jezdez: This was needed to have the volumes mounted in the worker and scheduler containers that previously didn't work.

bin/run
```diff
@@ -32,11 +32,11 @@ wait_for() {
 
 case $1 in
   web)
-    newrelic-admin run-python manage.py migrate
+    newrelic-admin run-python manage.py migrate --noinput
     exec newrelic-admin run-program gunicorn atmo.wsgi:application -b 0.0.0.0:${PORT} --workers 4 --access-logfile -
     ;;
   web-dev)
-    python manage.py migrate
+    python manage.py migrate --noinput
```
jezdez: I had a permission that was added in a different branch and not having this `--no-input` here basically broke the run of the web contains.

atmo/jobs/models.py
```diff
@@ -162,14 +177,11 @@ def should_run(self):
         active = self.start_date <= now
         if self.end_date is not None:
             active = active and self.end_date >= now
-        if not self.latest_run or self.latest_run.scheduled_date is None:
```
jezdez: Just some light refactoring..

atmo/jobs/jobs.py
```diff
@@ -1,30 +1,65 @@
...
-        # and then check if the job is expired and terminate it if needed
-        if job.is_expired:
-            # This shouldn't be required as we set a timeout in the bootstrap script,
-            # but let's keep it as a guard.
-            job.terminate()
+    jobs = SparkJob.objects.all()
+
+    with transaction.atomic():
+        # get the jobs with prior runs
+        jobs_with_runs = jobs.filter(runs__isnull=False).prefetch_related('runs')
```
maurodoglio: so here we are fetching all the jobs with at least a run, no matter the status of the run, right? Shouldn't we fetch only those with at least a run in a non final status?

docker-compose.yml
maurodoglio: 🎉

atmo/jobs/jobs.py
```diff
@@ -1,30 +1,65 @@
...
-            run_jobs.append(job.identifier)
-
-        # and then check if the job is expired and terminate it if needed
-        if job.is_expired:
-            # This shouldn't be required as we set a timeout in the bootstrap script,
-            # but let's keep it as a guard.
-            job.terminate()
+    jobs = SparkJob.objects.all()
+
+    with transaction.atomic():
```
maurodoglio: We could have a transaction for each job update, right? I don't see any advantage in using a single transaction for all the updates.

jezdez: the only advantage is to have it visible in the code that this is a separate transaction, but yeah, let's do a transaction per job

atmo/jobs/jobs.py
jezdez: good idea!

jezdez:
@maurodoglio Can you take another look at the last commits, just don't want to eff up the query..

atmo/jobs/jobs.py
```diff
@@ -21,17 +22,20 @@ def run_jobs():
     jobs = SparkJob.objects.all()
 
     # get the jobs with prior runs
-    jobs_with_runs = jobs.filter(runs__isnull=False).prefetch_related('runs')
-    logger.debug('Updating Spark jobs: %s', jobs_with_runs)
+    jobs_with_active_runs = jobs.filter(
+        runs__isnull=False,
```
maurodoglio: You can probably leave this condition out because of the second one.

atmo/jobs/jobs.py
```diff
@@ -1,30 +1,69 @@
...
+        job.latest_run.jobflow_id: job
+        for job in jobs_with_active_runs
+    }
+    # get the created dates of the job runs to limit the ListCluster API call
+    provisioner = ClusterProvisioner()
+    runs_created_at = jobs_with_active_runs.datetimes('runs__created_at', 'day')
+    logger.debug('Fetching clusters older than %s', runs_created_at[0])
+
+    cluster_list = provisioner.list(created_after=runs_created_at[0])
+    logger.debug('Cluster found: %s', cluster_list)
```
robhudson: "Clusters found"? This should return a list right?

jezdez: Yeah, "clusters"

atmo/jobs/jobs.py
```diff
@@ -1,30 +1,69 @@
...
+
+    for cluster_info in cluster_list:
+        # filter out the clusters that don't relate to the job run ids
+        job = jobflow_job_map.get(cluster_info['jobflow_id'], None)
```
robhudson: `None` is the default, not really needed.

atmo/jobs/jobs.py
```diff
@@ -1,30 +1,69 @@
...
+                job.run()
+                run_jobs.append(job.identifier)
+
+            # and then check if the job is expired and terminate it if needed
+            if job.is_expired:
+                logger.debug('Job %s is expired and is terminated', job)
+                # This shouldn't be required as we set a timeout in the bootstrap script,
+                # but let's keep it as a guard.
+                job.terminate()
     return run_jobs
```
robhudson: This is a huge improvement. Well done. 👍

Issue #213, opened 2017-02-13
Display notification when cluster is ready


jezdez:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1324464

marco-c:
It would be nice to use web-push here, there's a Python library at https://github.com/web-push-libs/pywebpush.

The README is short, but exhaustive. If you want to use it and you have questions, you can ask @jrconlin or me either here or on IRC.

jezdez:
Yep, there is an easy to use Django app that uses pywebpush called [django-webpush](https://github.com/safwanrahman/django-webpush) that I looked at before. It's neat and I think it's just a matter of sitting down defining what notifications we want and then hooking it up. No need to reinvent the wheel. @safwanrahman has been a Mozilla contributor for a while and has been super helpful as well.

rafrombrc:
Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1316633.

jezdez:
Since the last refactor of the cluster status system we can much easier now track the status, so tackling this would be much easier now.

PR #291, opened 2017-03-07
Alert emails and more

jezdez:


atmo/jobs/migrations/0013_sparkjobrunalert.py
```diff
@@ -0,0 +1,31 @@
...
+                ('reason', models.CharField(blank=True, help_text='The reason for the creation of the alert.', max_length=50, null=True)),
+                ('mail_sent_date', models.DateTimeField(blank=True, help_text='The datetime the alert email was sent.', null=True)),
+            ],
+            options={
+                'abstract': False,
+                'get_latest_by': 'modified_at',
+                'ordering': ('-modified_at', '-created_at'),
+            },
+        ),
+    ]
```
robhudson: I see the migration but I don't see the model in this commit?

atmo/models.py
```diff
@@ -130,3 +131,15 @@ def next_field_value(model_cls, field_name, field_value,
...
+class ForgivingReverseOneToOneDescriptor(ReverseOneToOneDescriptor):
+    def __get__(self, *args, **kwargs):
+        try:
+            return super().__get__(*args, **kwargs)
+        except self.RelatedObjectDoesNotExist:
+            return None
+
+
+class ForgivingOneToOneField(models.OneToOneField):
+    related_accessor_class = ForgivingReverseOneToOneDescriptor
```
robhudson: This has always been a frustration with `OneToOneField`s. 👍

atmo/jobs/migrations/0013_sparkjobrunalert.py
jezdez: It's in https://github.com/mozilla/telemetry-analysis-service/pull/291/files#diff-8d0f4fb6729f2067a211fb64aa9e32bc, did I add it in a later commit maybe?

jezdez: Hm, nah, it's all in https://github.com/mozilla/telemetry-analysis-service/pull/291/commits/d4c694ef511791148f53f78ccbe39ccdebd357a9

jezdez: Wait, the migration was added in https://github.com/mozilla/telemetry-analysis-service/commit/352de45bfd9acb2effd11f91a4a7da235be92d5a. Must be a rebase issue, reshuffling things...

atmo/models.py
jezdez: Yeah, me too, this is much nicer.

atmo/jobs/models.py
```diff
@@ -294,5 +294,35 @@ def update_status(self, info=None):
...
+        related_name='alert',  # run.alert & alert.run
+        primary_key=True,
+    )
+    reason = models.CharField(
+        max_length=50,
+        blank=True,
+        null=True,
+        help_text="The reason for the creation of the alert.",
+    )
+    mail_sent_date = models.DateTimeField(
```
maurodoglio: Can we call this field `notification_sent_date` ? The rest of the model doesn't specify the actual medium used to send the notification, and I really like it!

atmo/clusters/models.py
```diff
@@ -150,6 +152,9 @@ def provisioner(self):
     def get_absolute_url(self):
         return reverse('clusters-detail', kwargs={'id': self.id})
 
+    def get_full_url(self):
+        return urljoin(settings.SITE_URL, self.get_absolute_url())
```
jezdez: I'm rather safe than sorry about stuff like this, in case the `SITE_URL` stops having a trailing slash for example, this is a safe bet. This also deduplicates slashes, which string.join wouldn't.

atmo/jobs/models.py
jezdez: Oh, that name is by design, since I'm guessing that we'll add a new medium like webpush (e.g. #213) I want this to be specific to the medium. In the future I see another field called `webpush_sent_date` for example.

maurodoglio: Oh I see, it makes sense now. And I guess adding a new Notification model would not be very KISS.

jezdez: Yeah, no reason at the moment to have a separate table I thought, dunno. We can always split it later if we really need to.

maurodoglio: :+1:

PR #301, opened 2017-03-10
Update whats new wording for EFS

fbertsch:


WHATSNEW.md
```diff
@@ -2,10 +2,12 @@
...
 
-Please be aware that the Spark clusters are now **persisting
-the home directory between launches**. Caution is advised with
-regard to long running scripts that write to disk, especially
-with regard to write speed and volume size.
+Good news everyone - files are now persisted between clusters!
+In fact, make changes to your own dotfiles in your home dir -
+they will live on between clusters as well.
+
+We will be limiting your directory size to 20GB, so don't go
```
jezdez: Ah, I didn't know we made that decision yet. The idea of the whatsnew section is to indicate the current state, not what's going to happen. Could we remove this?

fbertsch: Seems kind of important that we let people know the size is limited. In that sense this is something that is part and parcel with our EFS usage, so in that sense *is* something that has already happened.

(( as an aside, we can't *really* limit the size, I just check it every once in awhile and email a bunch of people if it's too big ))

jezdez: @fbertsch Ah, understood, makes sense.

PR #318, opened 2017-03-16
Use Celery and further stabilize tasks

### LINKED ISSUE #317: Use Celery
To fix some of the issues (e.g. AWS API throtteling) and implement more advanced patterns of task processing (non-trivial task retries, depdenencies, task groups) we need to move to Celery. This will also provide the basis to move away from the naive implementation of a cron-like job queue, to one that schedules job runs without constant checking.

jezdez:
This does a few things:

- Move from RQ to Celery for better control of the queuing process and scheduling and processing guarantees.
- Implement jittered exponential retries (see https://www.awsarchitectureblog.com/2015/03/backoff.html and https://github.com/awslabs/aws-arch-backoff-simulator for more infos).
- Trigger some tasks only on database commit.
- Use more transaction wrapping.
- Uses [RedBeat](https://github.com/sibson/redbeat) for ensuring single beat in Celery cluster and handling scheduling.

Closes #317


atmo/clusters/tasks.py
```diff
@@ -94,21 +104,28 @@ def update_clusters_info():
...
+            cluster.most_recent_status = info['state']
+            cluster.save()
+
+            updated_clusters.append(cluster.identifier)
+
+            # if not given enqueue a job to update the public IP address
+            if not cluster.master_address:
+                transaction.on_commit(
+                    lambda: update_master_address.delay(cluster.id)
+                )
```
robhudson: Would it make sense to only enqueue this if the state is "RUNNING" (or other statuses)? E.g. if the state went from "STARTING" to "BOOTSTRAPPING", then this is a wasted API call b/c the IP address won't be there yet.

atmo/jobs/tasks.py
```diff
@@ -18,6 +19,7 @@
 logger = logging.getLogger(__name__)
 
 
+@celery.autoretry_task()
 @newrelic.agent.background_task(group='RQ')
```
robhudson: `group='Celery'`

atmo/jobs/tasks.py
```diff
@@ -75,26 +77,30 @@ def run_jobs():
     return run_jobs
 
 
+@celery.task
 @newrelic.agent.background_task(group='RQ')
```
robhudson: `group='Celery'`

atmo/urls.py
```diff
@@ -16,7 +16,6 @@
 
 urlpatterns = [
     url(r'^$', views.dashboard, name='dashboard'),
-    url(r'^admin/rq/', include('django_rq.urls')),
```
robhudson: Does the celery result backend have an admin we might want to enable?

atmo/clusters/tasks.py
jezdez: That's a great idea, will do!

atmo/urls.py
jezdez: It automatically registers itself with the admin like a regular Django app.

atmo/celery.py
```diff
@@ -0,0 +1,72 @@
...
+#   should have a `CELERY_` prefix.
+celery.config_from_object('django.conf:settings', namespace='CELERY')
+
+# Load task modules from all registered Django celery configs.
+celery.autodiscover_tasks()
+
+
+@celery.task(bind=True)
+def debug_task(self):
+    print('Request: {0!r}'.format(self.request))
```
jezdez: Nope, good catch.

robhudson:
r+ to the new changes.

atmo/clusters/tasks.py
```diff
@@ -3,26 +3,32 @@
...
 from .. import email
+from ..celery import celery
 from .models import Cluster
 from .provisioners import ClusterProvisioner
 
 
-@newrelic.agent.background_task(group='RQ')
-def delete_clusters():
+@celery.task
+@newrelic.agent.background_task(group='Celery')
```
maurodoglio: Isn't celery already instrumented on newrelic?

atmo/clusters/tasks.py
```diff
@@ -109,6 +118,11 @@ def update_clusters_info():
         cluster.most_recent_status = info['state']
         cluster.save()
 
+        updated_clusters.append(cluster.identifier)
+
         # if not given enqueue a job to update the public IP address
         if not cluster.master_address:
-            django_rq.enqueue(update_master_address, cluster.id)
+            transaction.on_commit(
```
maurodoglio: I didn't know about this! Nice!

atmo/settings.py
```diff
@@ -48,8 +48,12 @@ class Celery:
     CELERY_TASK_SEND_SENT_EVENT = True
```
maurodoglio: s/celery-beat/redbeat in the commit message

atmo/celery.py
```diff
@@ -11,13 +11,38 @@
 configurations.setup()
```
maurodoglio: It would be great to have the implementation of the backoff in a single commit. At the moment it's introduced in the first commit and fixed in this one. I'll leave it to you to decide whether it's worth it or not

atmo/clusters/tasks.py
jezdez: Ah, good point!

atmo/celery.py
jezdez: Sure, I can squash them

PR #324, opened 2017-03-17
Fix #283: Show all spark jobs to admins

### LINKED ISSUE #283: Show scheduled jobs of all users to admins
This is a split-off from #271 and basically would allow admins to see all scheduled jobs on the dashboard (and their last run's status) using the guardian permission system. Seeing details about the jobs can be seen via #271 on the job detail page.

robhudson:


robhudson:
I'm not sure if this is as simple as was expected? Or if we want to separate the admin user's own jobs vs the rest of the jobs? Similar to clusters there could be a button to filter: `[ My jobs | All ]`

jezdez:
@robhudson Ah, yeah, exactly, having an additional button filter that differentiates between "my jobs" (which in theory would include jobs that are "shared" by other users in a future feature) and an "all jobs" for members of the "spark job maintainers" group or simply all superusers is what I was looking for.

That group exists in prod already, but this is a good opportunity to decide which levels of access restrictions we want to maintain.

django-guardian currently provides us with per-object permissions for each job and cluster. That could be used later on for users to grant access to jobs or clusters to other users for shared work purposes.
On the administration side, I'm not a fan to use the blank "superuser" role as that is hard to take back in case we face new use cases in the future. What do you think?

robhudson:
The latest commits update this PR to use django-guardian's per object permissions allowing users in the 'Spark job maintainers' group to see all spark jobs via a separate tab navigation on the dashboard.

An example:
screenshot 2017-03-21 17 10 52

atmo/apps.py
```diff
@@ -79,6 +79,9 @@ def ready(self):
         # https://github.com/mozilla/sugardough/issues/38
         session_csrf.monkeypatch()
 
+        # Import job signals so they are registered.
+        import atmo.jobs.signals  # noqa
```
jezdez: I know what you're doing here, but I'd rather not use imports as triggers for signal connections. Can you keep the signal callbacks in atmo.jobs.signals and connect them here instead?

atmo/jobs/migrations/0016_assign_group_view_perm.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.contrib.auth.models import Group
+from django.db import migrations
+from guardian.shortcuts import assign_perm, remove_perm
+
+
+def assign_spark_job_view_permission_to_group(apps, schema_editor):
+    SparkJob = apps.get_model('jobs', 'SparkJob')
+    group, _ = Group.objects.get_or_create(name='Spark job maintainers')
+    assign_perm('view_sparkjob', group, SparkJob.objects.all())
```
jezdez: The last time I did that assign_perm did not work correctly inside the Python migration step. See the `PermissionMigrator` class in `atmo/models.py` as an alternative, which also makes sure the permission exists at the time of the migration run. Could you modify the class to offer a way to assign permissions to groups in addition to users and use that here please?

atmo/templates/atmo/dashboard.html
```diff
@@ -60,35 +60,24 @@
...
+        <li role="presentation"><a href="#all-jobs" aria-controls="all-jobs" role="tab" data-toggle="tab">All jobs</a></li>
+      </ul>
+    {% endif %}
+    <div class="tab-content">
+      <div class="tab-pane active" id="my-jobs" role="tabpanel">
+        {% include "atmo/_jobs_table.html" with spark_jobs=user_spark_jobs %}
+      </div>
+      {% if is_sparkjob_maintainer %}
+        <div class="tab-pane" id="all-jobs" role="tabpanel">
+          {% include "atmo/_jobs_table.html" with spark_jobs=group_spark_jobs %}
```
jezdez: Hm, I'd rather not use an include here, but have the view pass a different set of jobs similar to the active/terminated/failed/all filters. That way we don't have to do the queries for all jobs all the time if we don't need them on the dashboard.

atmo/views.py
```diff
@@ -35,28 +36,50 @@ def dashboard(request):
...
         use_groups=False,
         with_superuser=False,
     )
 
-    # a list of modification datetimes of the clusters and Spark jobs to use for getting the
-    # last changes on the dashboard
+    try:
+        group = Group.objects.get(name='Spark job maintainers')
+    except Group.DoesNotExist:
+        group = None
```
jezdez: `group = Group.objects.filter(name='Spark job maintainers').first()` suffices here, it'll fallback to `None` automatically and won't require an extra try/except clause.

tests/test_dashboard.py
```diff
@@ -12,31 +13,56 @@
...
 
-@pytest.fixture
-def dashboard_spark_jobs(now, test_user):
-    for x in range(10):
-        identifier = 'test-spark-job-%s' % x
+LAST_SPARK_JOB_ID = 0
+
+
+def create_spark_jobs(num, date, owner):
+    global LAST_SPARK_JOB_ID
```
jezdez: Hm, not a fan of this global, since we're not checking for the identifier to have this id (useless code from before) let's just use a random value here instead and get rid of the global variable.

atmo/templates/atmo/dashboard.html
jezdez: To clarify, the view would continue to just pass in `spark_jobs` but depending on a GET parameter `jobs` the value would be different. E.g. `/?jobs=all` would mean the `spark_jobs` variable would make the query for all user jobs.

atmo/apps.py
robhudson: IOW, don't use the decorator syntax but actually call `.connect` here? Sure, sounds better.

robhudson:
Re: "let's just use the same style of button group for the (My jobs | All users) like for the active/terminate/.. filters." ...

I asked in the channel yesterday why this UI was chosen. IMO it's not as good of a UI element to use for this purpose and is confusing to me, personally. I'm happy to make the change to fit the project, but would like to know if any thought went into the button groups vs the tabbed navigation for this need.

atmo/apps.py
jezdez: Cool, thanks 👍

jezdez:
@robhudson Yeah, that was after I left for the day, apologies.

I'd appreciate if you'd change this for here since I definitely put some thoughts into this when I switched it the tabs to button groups: 1) they scale better on responsive designs and 2) are the appropriate actionable UI elements for things that change the data that make up the UI.

E.g. a pagination UI would use the same button group style to differ between the various options to display the data on the page.

tests/test_dashboard.py
robhudson: It's still relevant since I'm calling `create_spark_jobs` twice and passing in two different users. Without it the spark job identifier will create clashes since it is required to be unique.

atmo/views.py
robhudson: I think the try/except is more explicit in the intention of the code. But I like the "one liner-ness" of using `.first()`. 👍

robhudson:
Question: Should I hide the "All" filter if the user isn't in the 'Spark job maintainers' group? Otherwise it will do nothing. Alternatively I could show it but make it inactive, which I don't think is best.

Best I think is to hide the filters and only show the user their jobs if they aren't in the group. If they are show the button group and allow them to filter by "All".

tests/test_dashboard.py
robhudson: Refactored it following the patterns in the cluster tests.

atmo/views.py
jezdez: Yeah, no problem with try/except blocks in general, I prefer them as well, but this is such a common case it was added to Django a while ago.

jezdez:
@robhudson Yep, hide the filter if the user doesn't have the group membership.

atmo/jobs/signals.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def assign_group_perm(sender, instance, created, **kwargs):
+    if created:
+        group, _ = Group.objects.get_or_create(name='Spark job maintainers')
+        assign_perm('jobs.view_sparkjob', group, instance)
+
+
+def remove_group_perm(sender, instance, **kwargs):
+        group, _ = Group.objects.get_or_create(name='Spark job maintainers')
```
jezdez: Indentation looks a bit off.

atmo/settings.py
```diff
@@ -394,6 +394,9 @@ class Core(AWS, Celery, Constance, CSP, Configuration):
                     'django.template.loaders.filesystem.Loader',
                     'django.template.loaders.app_directories.Loader',
                 ],
+                'libraries': {
+                    'atmo_tags': 'atmo.templatetags',
```
jezdez: Nice, I love that we don't have to put the template tags in submodules anymore. Could you rename the label though, so we can do `{% load atmo %}` in the templates instead the duplicated meaning `{% load atmo_tags %}`?

atmo/templatetags.py
```diff
@@ -0,0 +1,13 @@
+from django import template
+from furl import furl
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def url_update(url, **kwargs):
+    f = furl(url)
```
jezdez: Minor nit, one character variable name

atmo/jobs/signals.py
robhudson: good catch, thanks. Surprised flake8 didn't catch that.

Issue #326, opened 2017-03-20
`top-signatures-correlations` job was not scheduled/run on 2017-03-19


marco-c:
My scheduled job (`top-signatures-correlations`) was not scheduled nor run on 2017-03-19.

marco-c:
See also #351.

Do you think #372 will fix these issues?

marco-c:
At the moment, the last run time is 2017/04/04 at 11:40 UTC (checked by @fbertsch). I've scheduled it twice after that, but it didn't run.
I've also tried to remove the scheduled job and recreate it, but it didn't run anyway.

jezdez:
@marco-c So I've done some digging yesterday and I think I found the underlying issue, basically the way ATMO currently works (unintentionally) let's the time window when the job is supposed to be executed slip due to our naive cron implementation (that uses the last successful run as the base datetime for running the next one, and not the saved start date) on top of the massive API throttling that Amazon does for our system.

We've requested a raise for the throttle limits but were told that it can't be raised. We've already implemented the strategy (exponential back-off retries) that AWS recommended to us. That means we have no other way out other than rethinking the way we schedule the jobs in the first place (e.g. #332, #372).

I've prioritized #372 now and will also solve #332 to reduce the chance that a system hiccup via API throttling prevents the job to execute.

rafrombrc:
#372 has been completed, target deployment on 2017/04/27, we'll come back and verify that it did indeed fix this issue after the deployment, more soon.

marco-c:
It was executed tonight.

rafrombrc:
Great, thanks! Closing this issue.

marco-c:
I have another scheduled job which is behaving strangely, `top-fennec-signatures-correlations`.
It looks like in the past few days (since April 27) it has been running twice a day, with the two runs really close to each other:
```
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170429233705.log.gz
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170429233422.log.gz
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170428233436.log.gz
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170428233405.log.gz
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170427233629.log.gz
top-fennec-signatures-correlations/logs/top-fennec-signatures-correlations.20170427233600.log.gz
[...]

PR #327, opened 2017-03-20
Make run alert reason message a TextField.

jezdez:
Ref #326.


jezdez:
@maurodoglio This is breaking cluster status updates right now in prod, which means we'll need to do a deploy asap. r?

atmo/jobs/models.py
```diff
@@ -346,10 +346,8 @@ class SparkJobRunAlert(EditedAtModel):
         null=True,
         help_text="The reason code for the creation of the alert.",
     )
-    reason_message = models.CharField(
-        max_length=50,
-        blank=True,
-        null=True,
+    reason_message = models.TextField(
```
maurodoglio: I guess the error was caused by reason messages with more than 50 characters?

jezdez: Yeah, didn't expect Amazon to send small novels :-/

Issue #218, opened 2017-02-13
Add EMR release 5.3.1 to atmo


jezdez:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1336983

jezdez:
Now that #330 is closed, I'll be closing this. We can add new EMR Releases via the Django admin now. Adding them implies testing them first of course.

Issue #203, opened 2017-02-13
Refactor EMR scripts


jezdez:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1312747

jezdez:
@vitillo Regarding the emr-bootstrap-spark repository, can you clarify:

- whether it should continue to exist or if I can merge its content completely into this repo?
- should the Ansible playbook kept for testing purposes?

My gut feel is that this basically should live under a "deploy" folder in the root of this repo, including the Ansible playbook. Then update the ATMO and Airflow code to get the shell scripts from there (maybe even from GitHub's raw file serving URL?).

Since we have some ideas to make the ATMO code agnostic and remove the Mozilla bits (#199) is there anything you want me to keep an eye on?

vitillo:
> whether it should continue to exist or if I can merge its content completely into this repo? 

I am generally OK with merging its content in this repo as long as it's easy to deploy changes to our Spark environments independently from the web-service if the need arises.  

> should the Ansible playbook kept for testing purposes?

To create or update an environment a parametrized cloudformation script has to be run. Ansible provides a convenient way to do so, among other things. Do you have something else in mind?

Ideally it should be easy to deploy a new test environment using a script (e.g. playbook) manually, so that Operations & Engineering can experiment with different configuration options and what not.

rafrombrc:
This is really more of a meta-bug, which should ultimately be represented by a handful of other issues breaking this down into multiple pieces. Component issues will be opened soon and this issue will be updated to refer to them as that happens.

PR #330, opened 2017-03-21
Make EMR release an own data model.

### LINKED ISSUE #215: Store the list of available emr releases in a database table
See https://bugzilla.mozilla.org/show_bug.cgi?id=1335109

jezdez:
Fix #215, #218. Refs #203.


atmo/clusters/views.py
```diff
@@ -11,14 +11,15 @@
...
 from .forms import NewClusterForm
-from .models import Cluster
+from .models import Cluster, EMRRelease
 
 
 @login_required
 def new_cluster(request):
     initial = {
-        'identifier': '{}-telemetry-analysis'.format(user_display(request.user)),
+        'identifier': '%s-telemetry-analysis' % user_display(request.user),
```
robhudson: I always thought `.format()` was preferred for Python 3?

atmo/jobs/forms.py
```diff
@@ -184,12 +186,6 @@ def __init__(self, *args, **kwargs):
 
     class Meta(BaseSparkJobForm.Meta):
         fields = BaseSparkJobForm.Meta.fields + ['emr_release']
```
robhudson: Is this line still required if you added the field to the class above?

atmo/clusters/views.py
jezdez: Hmyeah not really, the old format wasn't removed from Python 3, so there is no reason to use it, especially since it has some spotty support in things like translations, so it's easier to just stick with one formatting. Plus there was [this thing recently](http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/) which made me just go back again.

atmo/jobs/forms.py
jezdez: Yeah, that's still required since while the form field is still rendered when not adding this to the field list, the form validation would not happen for it since the form class inherits the form meta class from `BaseSparkJobForm`, which doesn't specify the `emr_release` field.

jezdez:
@robhudson I'd rather not provide a guide for adding new EMR versions since a) those need to be tested with our EMR bootstrap script so adding new ones need to be carefully considered b) the PR adds the currently active EMR versions as a data migration, which is enough for now

atmo/clusters/models.py
```diff
@@ -9,10 +9,79 @@
...
+        'of software like Hadoop, Spark, etc. '
+        'See <a href="'
+        'http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-whatsnew.html"'
+        '>what\'s new</a> in each.'
+    )
+    emr_release = models.ForeignKey(
+        EMRRelease,
+        verbose_name='EMR release',
+        on_delete=models.PROTECT,
+        related_name='created_%(class)ss',  # e.g. emr_release.created_clusters.all()
```
maurodoglio: Nice! I didn't know you can do that.

atmo/clusters/models.py
```diff
@@ -9,10 +9,79 @@
...
+        return self.filter(is_experimental=False, is_deprecated=False)
+
+    def experimental(self):
+        return self.filter(is_experimental=True)
+
+    def deprecated(self):
+        return self.filter(is_deprecated=True)
+
+
+class EMRRelease(EditedAtModel):
```
maurodoglio: What about adding a `is_active` attribute? is_active==False would be the final state of a EMRRelease object.

jezdez: Good idea!

atmo/clusters/models.py
jezdez: Yeah, that's useful to have APIs that are easier to use.

PR #339, opened 2017-03-24
Refactor autoretrying as separate decorator and use it with ClientError.

### LINKED ISSUE #336: Handle botocore.exceptions.ClientError exceptions in Celery tasks gracefully
It seems as if botocore's magic exception prevents Celery from correctly storing them in the result backend. Best would be to wrap them in a custom decorator and raising them again with an own exception class.

jezdez:
Fix #336.

### DIFF
--- a/atmo/celery.py
+++ b/atmo/celery.py
@@ -3,6 +3,7 @@ import os
 import random
 from celery import Celery
 from celery.five import string_t
+from celery.task import current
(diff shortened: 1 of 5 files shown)


jezdez:
@robhudson I'll add a test, you're right. While the infra code for this is tested in Celery, we should still make sure we have the custom bits covered.

PR #355, opened 2017-03-30
Add lifetime to cluster

### LINKED ISSUE #298: Reduce the default cluster lifetime to 8h
I believe most of the people use clusters during working hours, so defaulting to 8h seems a good approximation. This will make us save some money for those cases where the owner simply forget about the running cluster. Of course this requires #221 to be solved first, to address those special cases where the user actually wants to run a 24h (or more) long analysis.

jezdez:
Fix #298.


atmo/clusters/models.py
```diff
@@ -248,7 +251,7 @@ def is_ready(self):
     @property
     def is_expiring_soon(self):
         """Returns true if the cluster is expiring in the next hour."""
-        return self.end_date <= timezone.now() + timedelta(hours=1)
+        return self.end_date <= timezone.now() + timedelta(hours=self.lifetime)
```
robhudson: If you change this I think the name is wrong. `is_expiring_soon` wouldn't make sense if `lifetime` was 8 hours away. That doesn't seem "soon" to me.

jezdez: Wait, this is a wrong change in the first place, wow!

jezdez: Thanks for catching this!

tests/test_clusters.py
```diff
@@ -108,6 +111,14 @@ def test_create_cluster(client, user, emr_release, ssh_key, cluster_provisioner_
 
 
 @pytest.mark.django_db
+def test_is_expiring_soon(cluster):
+    assert not cluster.is_expiring_soon
+    cluster.end_date = timezone.now() + timedelta(minutes=59)  # the cut-off is at 1 hour
+    cluster.save()
+    assert cluster.is_expiring_soon
```
robhudson: 👍 💣

Issue #351, opened 2017-03-30
Job scheduled to run at 10 AM was scheduled and run at 1 AM


marco-c:
I've scheduled my `top-signatures-correlations` job to run at 10 AM. Today it was scheduled and run at 1 AM:
```
Last scheduled date
2017-03-30 01:15
Last run date
2017-03-30 02:15
```

marco-c:
Notice it was scheduled and run at 1 AM today (2017-03-30), but skipped 10 AM yesterday (2017-03-30).

marco-c:
Today:
Start date - 2017-04-04 17:00
Last scheduled date - 2017-04-04 11:40
Last run date - n/a
Last terminated date - 2017-04-04 22:00

I've got an email at 22:00 UTC saying that the job at 11:40 UTC failed.

I'm not sure what happened to the 17 UTC job.

jezdez:
@marco-c While investigating #326 I've found that ATMO's naive cron implementation mishandles the start date which leads to the start time of consecutive runs to slip slowly as it depends on previous runs instead on the originally saved start date. That's why your assumption of the job to always start at the start time can't currently be guaranteed. This is a bug and I'll fix it at the same time when I work on #372.

jezdez:
This should be fixed since #379 was merged.

Issue #255, opened 2017-02-23
Check for race conditions when creating clusters


jezdez:
We've had two cases in which we speculate a throttling issue caused newly created Spark clusters to not be stored in the atmo database and live on in EMR without checking. 

We should check how we can make sure the clusters are first stored in the database (including by manually checking for committed database transactions, not only relying on the auto-commit behavior) and then launch the clusters after the fact.

jezdez:
#318 contains changes that adds more transaction blocks in the tasks that create and update clusters. That's important when we make calls to the AWS API which may fail due to throttling and result in broken transactions.

Issue #190, opened 2017-02-13
Scheduled jobs should have an option to be run "right now"


jezdez:
https://bugzilla.mozilla.org/show_bug.cgi?id=1309567

PR #379, opened 2017-04-07
Replace naive cron-like job runner with native scheduler

### LINKED ISSUE #372: Use Celery scheduler for running jobs periodically
We've recently moved to Celery and now can use [redbeat](https://github.com/sibson/redbeat) to properly schedule tasks using Celery's scheduler instead of using a pseudo cron that checks every 5 minutes if it should schedule tasks. This basically removes our naive implementation with one that is not naive.

jezdez:
Fix #372 #351 #255 

Adds a Celery task for running a Spark job.

This task is used of Redbeat to schedule the Spark jobs using the Celery beat.
We add/remove Spark jobs from the schedule on save/delete and can restore the schedule from the database again.
This will also be useful for adding a “run now” button to Spark jobs.

This does a bunch of other things:

- send expiration emails for Spark jobs
- split tests into clusters and jobs packages with topic submodules (test_models, test_provisioners, test_views etc)
- refactor some test fixtures (e.g. remove all the pytest.mark.django_db decorators in favor of an autoused fixture)
- extend Celery task tests dramatically
- extend the test factories to cater to new tests
- fixed some model manager methods and test them
- refactor django.contrib.messages testing
- extend tests in various corner cases

Refs #190 and possibly others


jezdez:
@robhudson So this is getting to a point at which I'm ready to not rebase commits again. Do you have some time to review this in between?

atmo/jobs/tasks.py
```diff
@@ -13,35 +13,85 @@
...
-@celery.task(max_retries=3)
-@celery.autoretry(ClientError)
-def run_jobs():
+@celery.task
+def send_expired_mails():
+    expired_spark_jobs = SparkJob.objects.filter(
+        expired_date__isnull=False,
+    )
+    for spark_job in expired_spark_jobs:
+        with transaction.atomic():
```
robhudson: Do you need `transaction.atomic` here? I don't see any database calls.

atmo/jobs/models.py
```diff
@@ -263,7 +263,13 @@ def terminate(self):
         if self.latest_run:
             self.cluster_provisioner.stop(self.latest_run.jobflow_id)
 
+    def run_now(self):
+        from .tasks import run_job
+        run_job.delay(self.pk)
+
     def save(self, *args, **kwargs):
+        # whether the job is being created for the first time
+        run_now = self.pk is None
```
robhudson: Does the comment match the code? Do we always want to run the job when saved? If I create a job to only run on weekends I wouldn't want my job to run if I save it on a Wednesday.

jezdez:
Uplifting the question by @robhudson [from the commit](https://github.com/mozilla/telemetry-analysis-service/pull/379/files/f3a63abeb251e2ebd47f5ddf04462997aa27d413#r112271735) to the PR:

> Does the comment match the code? Do we always want to run the job when saved? If I create a job to only run on weekends I wouldn't want my job to run if I save it on a Wednesday.

@robhudson The condition here for running the job is having no primary key yet, which happens when the `SparkJob.save` method is being loading for the first time, in which case there is no row in the table yet. I know it's not ideal but I assumed that's closest to the intention of the user from running a job right away when scheduling a Spark job.

As a reminder, the ATMO workflow is that when you schedule a new Spark job (e.g. https://atmo.stage.mozaws.net/jobs/new/), you're selecting a start date (and optionally an end date). That's the basis for when ATMO should start counting for when the next run happens. It's repeated from that point on by the selected "run interval".

The help text for the start date field says: "Date and time of when the scheduled Spark job should start running." So when we save the Spark job we need to somehow make sure that it's being scheduled at the time that the user expects (start_date) and then repeated.

The Celery scheduling API allows providing a time to start (by default "now") and then the timedelta at which the task run should be repeated. The problem is that if we were to use the saved start_date as the start date for the schedule, it will only

jezdez:
@robhudson I've updated the code for the first run of a Spark job a bit more clear in ec3005daafe24e6362fd3409a69ccc58703f5b88.

- We use the "start_date" field as the starting point for when Celery should count intervals.
- Celery won't actual run the job right away but only when one interval has passed, which is in our case at least a day.
- Instead we're scheduling the first run of a Spark job "manually" using the `eta` parameter with the `start_date` value so even if that's in the future, Celery will pick it up and also will start counting the intervals from that point on.

atmo/jobs/models.py
jezdez: Replied in the PR instead so we don't lose this discussion in a commit.

atmo/jobs/tasks.py
jezdez: Hm, nope you're right. Will remove.

atmo/jobs/models.py
```diff
@@ -178,50 +208,25 @@ def is_public(self):
...
-        now = timezone.now()
-        active = self.start_date <= now
-        if self.end_date is not None:
-            active = active and self.end_date >= now
-        return (
-            self.is_enabled and
-            active and
-            self.is_due(now)
-        )
+        return self.is_runnable() and self.is_enabled and self.is_due()
```
jezdez: I was a bit confused by the amount of logic behind those properties and thought methods would better match that. Also `is_enabled` is a model field, which makes it confusing if all `is_*` members are properties. Do you prefer if I were to turn `is_due` and `is_runnable` into properties again?

robhudson: I'd go with your preference. I just wanted to point it out. Any future development would really require looking those up to know if it's a method or property. Since these are intended to be booleans, if used improperly, it could do the totally wrong thing...

```
>>> def is_something():
...     return False
...
>>> bool(is_something)
True
```

jezdez: I've replaced the methods again with properties, you're right. One thing to note is that when mocking those properties, the pattern is:

```python
def test_foo(mocker):
    mocker.patch(obj, 'is_bar', new_callable=mocker.PropertyMock, return_value=True)
    assert obj.is_bar
```

atmo/jobs/tasks.py
```diff
@@ -13,67 +13,215 @@
...
     """
-    Run all the scheduled tasks that are supposed to run.
+    Periodic task to purge all schedule entries
+    that don't have a SparkJob instance anymore
     """
-    # first let's update the job statuses if there are prior runs
-    run_jobs = []
-    jobs = SparkJob.objects.all()
+    expired_spark_jobs = []
+    for spark_job in SparkJob.objects.filter(end_date__lte=timezone.now()):
```
jezdez: Yeah, good idea, fixed in eed45a6f246590235670d78815187abd466748c6

jezdez:
@robhudson Nah, I think we'll see how this goes once it's on stage.

PR #383, opened 2017-04-10
correct logic on existance of .env file

peterbe:
If the file does *not* exist, *then* it should copy from the dist version. Right?

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,7 +2,7 @@
 set -eo pipefail
 
 # copy the .env template to .env if not there already
-[ ! -f .env ] && cp .env-dist .env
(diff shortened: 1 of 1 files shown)


jezdez:
Hm, I don't get this, the condition is `[ ! -f .env ]` returns `True` if the file **doesn't** exist. So && is correct here. You can easily try this out in a shell with `[ ! -f testfile.txt ] && touch testfile.txt`. Or am I missing something?

peterbe:
Ugh. I blame Mondays. 
To test myself I copied the line into a `/tmp/test.sh` and ran it to check for reals. Turns out I copied that wrong :(

PR #398, opened 2017-04-25
Email fixes, send email when Spark job times out.

jezdez:


atmo/templates/atmo/jobs/mails/failed_run_alert.mail
```diff
@@ -1,4 +1,13 @@
-{% load atmo %}Your scheduled Spark job "{{ alert.run.spark_job.identifier }}" has failed
+{% load atmo %}
```
robhudson: Is whitespace no longer a concern with `mail_builder`?

jezdez: Nope, it strips [subject and body and the other fields](https://github.com/funkybob/django-mail-builder/blob/b4523fa41d9579552cda61e1c5d5b3d6d4fda825/mail_builder/builder.py#L25-L40)

PR #400, opened 2017-04-26
Fix #294 - Use unique username from email address instead of first name for user username.

### LINKED ISSUE #294: Rename usernames from <firstname> to <flastname>
When the allauth backend was added I missed the chance to use the <flastname> format that Mozilla uses for email addresses as the username. Instead it currently uses the firstname of the user as the username which has some hilarious side effect for common names.

jezdez:


atmo/users/adapters.py
```diff
@@ -33,3 +35,11 @@ def validate_disconnect(self, account, accounts):
...
                 'At least one account needs to be connected'
             )
+
+    def populate_user(self, request, sociallogin, data):
+        user = super().populate_user(request, sociallogin, data)
+        email = user_email(user)
+        if email and '@' in email:
+            username = email.split('@')[0]
+            user_username(user, username)
+        return user
```
robhudson: Not sure where or how this is used. Do you need to do a `user.save()` here since you're setting the username attribute?

atmo/users/migrations/0002_rewrite_usernames.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.13 on 2017-04-26 09:39
+from __future__ import unicode_literals
+
+from django.db import migrations
+from allauth.utils import generate_unique_username
+from allauth.account.utils import user_username, user_email, user_field
```
robhudson: account > utils

atmo/users/migrations/0002_rewrite_usernames.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def fix_usernames(apps, schema_editor):
+    User = apps.get_model('auth', 'User')
+    all_usernames = User.objects.all().values_list('username', flat=True)
```
robhudson: `all_usernames` doesn't appear to be used?

atmo/users/adapters.py
jezdez: Nah, this method is only about [populating the various attributes](http://django-allauth.readthedocs.io/en/latest/advanced.html#creating-and-populating-user-instances). It's used as part of the signup/login flow in allauth. [Here's the code](https://github.com/pennersr/django-allauth/blob/55e12f9ec1f83f82a2013a9a840ab71e596c0aac/allauth/socialaccount/adapter.py#L84-L112). The `user_username` helper is a generic getter/setter.

atmo/users/migrations/0002_rewrite_usernames.py
jezdez: Good catch, that's a leftover from when I wasn't using `generate_unique_username` which does this behind the scenes.

PR #401, opened 2017-04-26
Fix #190 - Add ability to run a job right now.

### LINKED ISSUE #190: Scheduled jobs should have an option to be run "right now"

jezdez:


mreid-moz:
Just driving by to say 👍

atmo/templates/atmo/jobs/run.html
```diff
@@ -0,0 +1,31 @@
...
+
+{% block page_title %}Run Spark job {{ spark_job }}{% endblock %}
+
+{% block content %}
+<div class="page-header">
+  <h2>Run Spark job <small>{{ spark_job }}</small></h2>
+</div>
+<div class="row">
+  <div class="col-sm-6">
+    <h4>Do you really want to run the Spark ?</h4>
```
robhudson: I want to run the spark! This sounds like a sci-fi movie I'm forgetting.

atmo/templates/atmo/jobs/delete.html
```diff
@@ -9,10 +9,7 @@
 </div>
 <div class="row">
   <div class="col-sm-6">
-    <h4>Do you really want to terminate the cluster?</h4>
```
robhudson: Good catch.

atmo/jobs/views.py
```diff
@@ -133,3 +136,42 @@ def download_spark_job(request, id):
     )
     response['Content-Length'] = spark_job.notebook_s3_object['ContentLength']
     return response
+
+
+@login_required
+@view_permission_required(SparkJob)
+def run_spark_job(request, id):
+    spark_job = SparkJob.objects.get(pk=id)
```
robhudson: Should this be a get or 404?

jezdez: yeah, it should, good catch.

atmo/templates/atmo/jobs/run.html
jezdez: LOL

atmo/jobs/views.py
jezdez: Ah, wait, the 404 is already handled by the view_permission_required decorator, which is why there is no need to do that. I should probably add an option to make it clear that this is what is happening.

jezdez:
The job detail page will look like this:
screen-shot-2017-04-27-15-17-41

PR #409, opened 2017-05-01
Fix #375: Add status and visual indicator of running jobs

### LINKED ISSUE #375: Strategize Spot Instance Usage
We are running into some problems with spot instances:

1. Our pricing (and limited machine type/region) is really killing the market. Pricing is up 2x at some points, and we plateau multiple times a day at our max ($.84). See https://pageshot.net/6A2aqHOaRiLpomfD/us-west-2.console.aws.amazon.com.

2. When pricing goes beyond ours, many of our machines get shut down, and lots of clusters are lost. This isn't as big of a deal with EFS, but it is still a time-waster and difficult to deal with every time, and that's an extra 15-30 minutes for N machines which we have to now pay to bootstrap again.

#229 is related to this.

robhudson:
screenshot 2017-05-01 14 24 01

### DIFF
--- a/atmo/jobs/models.py
+++ b/atmo/jobs/models.py
@@ -385,6 +385,15 @@ class SparkJobRun(EditedAtModel):
     def info(self):
         return self.spark_job.cluster_provisioner.info(self.jobflow_id)
 
+    @property
+    def status_css_class(self):
+        if self.status in Cluster.ACTIVE_STATUS_LIST:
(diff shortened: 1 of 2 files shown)


robhudson:
Pretty simple change. Let me know if this aligns with your vision or if there's more to do.

robhudson:
Got the issue number incorrect. Closing

PR #410, opened 2017-05-01
Fix #374: Add status and visual indicator of running jobs

### LINKED ISSUE #374: Add a status indicator that shows if a scheduled job is currently running

robhudson:


atmo/jobs/models.py
```diff
@@ -385,6 +385,15 @@ def spark_job_identifier(self):
...
         return self.spark_job.cluster_provisioner.info(self.jobflow_id)
 
+    @property
+    def status_css_class(self):
+        if self.status in Cluster.ACTIVE_STATUS_LIST:
+            return 'success'
+        if self.status in Cluster.FAILED_STATUS_LIST:
+            return 'danger'
+
+        return None
```
jezdez: A few notes here:
- There are already helper properties called `Cluster.is_active` and `Cluster.is_failed` that are in line with this logic and we should add similar helpers to the `SparkJobRun` model
- There is no need to have the CSS class thing happening in the data model, it's mixing data handling with representation which will be harder to maintain on the long run, let's move this into a template tag

atmo/templates/atmo/dashboard.html
```diff
@@ -83,11 +83,12 @@
           <th>Job timeout</th>
           <th>Start date</th>
           <th>End date</th>
+          <th>Status</th>
         </tr>
       </thead>
       <tbody>
         {% for spark_job in spark_jobs %}
-        <tr>
+        <tr class="{{ spark_job.latest_run.status_css_class }}">
```
jezdez: I'm not a fan of using colored rows for this since it can at worst become a rainbow like table that is hard to read. Instead let's use bold, colored text for the status text maybe a descriptive icon from the [Glyphicons](http://getbootstrap.com/components/#glyphicons) set bootstrap provides. E.g. "glyphicon glyphicon-ok" for succeess, "glyphicon glyphicon-remove" for error?

robhudson:
Updated. New screenshots attached...

screenshot 2017-05-03 15 20 15
screenshot 2017-05-03 15 20 09

PR #421, opened 2017-05-05
Make sure the scheduler will only ever run run as a singleton.

### LINKED ISSUE #417: Job kicked off twice yesterday
I got a failure email last night for my txp-install-uninstall-events job but looking at the cluster/job list it looks like it's because the job was queued twice (and both ran at the same time.)

jezdez:
This uses a fix for redbeat’s distributed Redis lock.
Removing some of the custom setting fixes we had to do as well.

Fix #417.

### DIFF
--- a/atmo/celery.py
+++ b/atmo/celery.py
@@ -8,8 +8,6 @@ from celery import Celery
 from celery.five import string_t
 from celery.task import current
 from celery.utils.time import maybe_iso8601
-from redbeat.schedulers import add_defaults
-from redis.client import StrictRedis
 
 
 # set the default Django settings module for the 'celery' program.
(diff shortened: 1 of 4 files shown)


jezdez:
To reproduce this you can call docker-compose scale to spin up another scheduler container.

In other shell run: `make up`
In other shell: `docker-compose scale scheduler=2`

You'll see the second scheduler boot up, but it won't actually do anything. If you want to see even better what's going on, change the loglevel of the scheduler command in the `bin/run` script to `-l debug`.

robhudson:
I tried building locally and got a hash error:
```
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    https://github.com/jezdez/redbeat/archive/lock-settings.zip (from -r requirements.txt (line 357)):
        Expected sha256 a08ed9b9fec47b115a424996109dd4febab84e25ebe1292a17f85f21e81e4489
             Got        19b1d6928eb5d6cee30fe950f5459675e093d5e4d1a862950099a6d64d740cf6
```

jezdez:
@robhudson The branch I worked in changed since I rebased and I forgot to update the PR here. The good news is that the PR merged in redbeat and I've just updated this PR to use the canonical repo as the source (albeit still not a release from PyPI).

PR #465, opened 2017-05-15
Update all job runs, not only the latest to prevent race conditions with throttling.

### LINKED ISSUE #455: Update all past job runs, not only latest
When updating the job run status we should make sure to update all past ones as well, to prevent race conditions and the time frame of the cluster query growing over time. /cc @robotblake

jezdez:
Fix #455

### DIFF
--- a/atmo/jobs/factories.py
+++ b/atmo/jobs/factories.py
@@ -36,7 +36,7 @@ class SparkJobRunFactory(factory.django.DjangoModelFactory):
     run_date = None
     terminated_date = None
     emr_release_version = factory.LazyAttribute(lambda run: run.spark_job.emr_release.version)
-    # created_at = factory.LazyFunction(timezone.now)
+    created_at = factory.LazyFunction(timezone.now)
 
     class Meta:
         model = models.SparkJobRun
--- a/atmo/jobs/factories.py
+++ b/atmo/jobs/factories.py
@@ -36,7 +36,7 @@ class SparkJobRunFactory(factory.django.DjangoModelFactory):
     run_date = None
     terminated_date = None
     emr_release_version = factory.LazyAttribute(lambda run: run.spark_job.emr_release.version)
-    created_at = factory.LazyFunction(timezone.now)
+    # created_at = factory.LazyFunction(timezone.now)
 
     class Meta:
         model = models.SparkJobRun
--- a/atmo/jobs/tasks.py
+++ b/atmo/jobs/tasks.py
@@ -84,7 +84,7 @@ def update_jobs_statuses():
     updated_spark_job_runs = []
     if runs_created_at:
         earliest_created_at = runs_created_at[0]
-        logger.debug('Fetching clusters older than %s', earliest_created_at)
+        logger.debug('Fetching clusters since %s', earliest_created_at)
 
         cluster_list = provisioner.list(created_after=earliest_created_at)
         logger.debug('Clusters found: %s', cluster_list)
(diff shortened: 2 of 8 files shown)


atmo/jobs/factories.py
```diff
@@ -36,7 +36,7 @@ class SparkJobRunFactory(factory.django.DjangoModelFactory):
     run_date = None
     terminated_date = None
     emr_release_version = factory.LazyAttribute(lambda run: run.spark_job.emr_release.version)
-    created_at = factory.LazyFunction(timezone.now)
+    # created_at = factory.LazyFunction(timezone.now)
```
robhudson: ❓

atmo/jobs/tasks.py
```diff
@@ -62,54 +62,49 @@ def expire_jobs():
...
-        'runs__created_at', 'day'
-    )
+    runs_created_at = active_spark_job_runs.datetimes('created_at', 'day')
 
     # only fetch a cluster list if there are any runs at all
     updated_spark_job_runs = []
     if runs_created_at:
-        logger.debug('Fetching clusters older than %s', runs_created_at[0])
+        earliest_created_at = runs_created_at[0]
+        logger.debug('Fetching clusters older than %s', earliest_created_at)
```
robhudson: It looks like this might be "Fetching clusters since %s". Older than would be this date and prior, yet the code below queries `created_after` this date.

tests/jobs/test_tasks.py
```diff
@@ -248,51 +248,76 @@ def test_update_jobs_statuses_empty(mocker, now, user):
...
         created_by=user,
-        run__status=Cluster.STATUS_RUNNING,
     )
-    spark_job1.latest_run.created_at = spark_job1.start_date
-    spark_job1.latest_run.save()
+    spark_job1_run = spark_job_run_factory.create(
+        spark_job=spark_job1,
+        status=Cluster.STATUS_RUNNING,
+    )
+    # setting created_at explicitely here since factoryboy isn't handling
```
robhudson: s/explicitely/explicitly/

tests/jobs/test_tasks.py
```diff
@@ -248,51 +248,76 @@ def test_update_jobs_statuses_empty(mocker, now, user):
...
+    # auto_now fields sch as created_at nicely, same for the other runs below
```
robhudson: s/sch/such/ ?

atmo/jobs/factories.py
jezdez: Ugh, sorry, yeah, debugging ..

atmo/jobs/tasks.py
jezdez: Good catch! 👍

atmo/settings.py
```diff
@@ -97,7 +97,7 @@ class Celery:
             },
         },
         'update_jobs_statuses': {
-            'schedule': crontab(minute='*/15'),
+            'schedule': crontab(minute='*'),
```
robotblake: Was this intentional?

jezdez: It was not. Sigh!

jezdez: Also, man, good catch!

atmo/jobs/tasks.py
```diff
@@ -62,54 +62,49 @@ def expire_jobs():
...
-                cluster_info['jobflow_id'],
-                None
-            )
-            if spark_job is None:
+            spark_job_run = spark_job_run_map.get(cluster_info['jobflow_id'])
+            if spark_job_run is None:
                 continue
+
             logger.debug(
                 'Updating job status for %s, latest run %s',
```
robotblake: Small thing but I wonder if we should remove "latest" from this log line.

jezdez: Good idea 👍

PR #474, opened 2017-05-18
Fix #271: Add job history details

### LINKED ISSUE #271: Create view with jobs histories
There have been several jobs failing silently. Those jobs will soon generate alerts (#201) but it would  still be convenient to have a master view in the dashboard that shows the history, and their status, of all scheduled jobs.  Furthermore, every user should be able to see the history for their own jobs.

robhudson:
screenshot 2017-05-17 17 06 51

### DIFF
--- a/atmo/templates/atmo/jobs/detail.html
+++ b/atmo/templates/atmo/jobs/detail.html
@@ -36,6 +36,7 @@
       <a href="#notebook" class="btn btn-sm btn-default active" aria-controls="notebook" role="tab" data-toggle="tab">Notebook</a>
       <a href="#results" class="btn btn-sm btn-default" aria-controls="results" role="tab" data-toggle="tab">Results</a>
       <a href="#logs" class="btn btn-sm btn-default" aria-controls="logs" role="tab" data-toggle="tab">Logs</a>
+      <a href="#runs" class="btn btn-sm btn-default" aria-controls="runs" role="tab" data-toggle="tab">Runs</a>
     </div>
     <div class="btn-group"
          role="group"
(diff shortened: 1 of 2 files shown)


robhudson:
What's in the table is basically what's in the spark job runs table.

Some questions:
* Do we need pagination when this table gets large?
* Are there other columns you can think of?

mreid-moz:
> Are there other columns you can think of?

If feasible, a link to the logs would be great. Also the cluster size, in case the user updates it.

robhudson:
> If feasible, a link to the logs would be great. Also the cluster size, in case the user updates it.

Right now we don't store the logs with the run data we have. The list of logs in the logs pane is simply a listing of the S3 bucket where the logs are stored. But, this could be changed so we can associated exactly which logs apply to which run. Since the logs have a date stamp in them it's probably pretty easy to map them together but I think it'd be nice to relate them on the backend as well.

As far as cluster size we're currently only tracking that at the job level, not for each run so we don't have that data. But again, we can change it on the backend if that's what we want to do.

robhudson:
I filed issues for the logs and cluster size.

mreid-moz:
Thanks! Both are just nice-to-haves, so a separate followup issue is fine w/ me.

atmo/templates/atmo/jobs/detail.html
```diff
@@ -163,6 +164,36 @@
...
+            </thead>
+            <tbody>
+              {% for run in spark_job.runs.all %}
+              <tr>
+                <td>{{ run.jobflow_id }}</a></td>
+                <td>{{ run.emr_release_version }}</td>
+                <td>{{ run.status }}</td>
+                <td>{{ run.scheduled_date|default:"n/a" }}</td>
+                <td>{{ run.run_date|default:"n/a" }}</td>
+                <td>{{ run.terminated_date|default:"n/a" }}</td>
```
jezdez: This will require an update when #481 is merged.

atmo/templates/atmo/jobs/detail.html
```diff
@@ -163,6 +164,36 @@
...
+                <th>Last status</th>
+                <th>Scheduled</th>
+                <th>Ran</th>
+                <th>Terminated</th>
+              </tr>
+            </thead>
+            <tbody>
+              {% for run in spark_job.runs.all %}
+              <tr>
+                <td>{{ run.jobflow_id }}</a></td>
```
jezdez: It'd be nice to use the Jobflow ID to link to the AWS EMR console if the current user has permissions (is staff? member of "Spark job maintainers" group which exists in prod?).

The URL for a job is `https://<AWS region>.console.aws.amazon.com/elasticmapreduce/home?region=<AWS region>#cluster-details:<Jobflow ID>`, e.g. `https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-IQROSTWRHXIH`.

robhudson: Is the AWS region always the same? Do we need to have any permissions for the link? The detail page is using the `@view_permission_required(SparkJob)` and the link requires AWS credentials so if they have those they are good.

robhudson:
Updated with the cluster size and link to AWS EMR console.

atmo/templates/atmo/jobs/detail.html
```diff
@@ -163,6 +164,38 @@
...
+                <th>Last status</th>
+                <th>Scheduled</th>
+                <th>Ran</th>
+                <th>Terminated</th>
+              </tr>
+            </thead>
+            <tbody>
+              {% for run in spark_job.runs.all %}
+              <tr>
+                <td><a href="https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:{{ run.jobflow_id }}">{{ run.jobflow_id }}</a></td>
```
jezdez: Let's not hard code the AWS region here, use this instead:

```
<td><a href="https://{{ settings.AWS_CONFIG.AWS_REGION }}.console.aws.amazon.com/elasticmapreduce/home?region={{ settings.AWS_CONFIG.AWS_REGION }}#cluster-details:{{ run.jobflow_id }}">{{ run.jobflow_id }}</a></td>
```

atmo/templates/atmo/jobs/detail.html
jezdez: @robhudson The region is configured in the `settings.AWS_CONFIG['AWS_REGION']` setting, so let's use that for the value. I made a comment in the review for this.

jezdez: Oh, and you're right, the permissions suffice.

Issue #264, opened 2017-02-27
When terminating a cluster, the end_date should be updated.


maurodoglio:
Looking at the atmo data on STMO I found out that almost all the clusters have the same duration (24h), so there's no way to know the exact amount of cpu hours used.

jezdez:
Hm, `end_date` defines the time until when the cluster should be automatically EOL'd, but we don't offer setting that via the UI, so it will by design always be 24 hours by default. I'm trying to understand what you try to do with the query on STMO, are you looking for the outliers that are *not* killed exactly after 24 hours?

maurodoglio:
A cluster can be either killed manually or by the scheduler. I think we should update the end_date in the former case, to reflect the actual termination date of the cluster. This will let me tell how long  each cluster has been running for, which is not possible at the moment.

rafrombrc:
This is best done by getting the info from the stream rather than the current state-based job status, blocked by #332.

rafrombrc:
Turns out #332 isn't a hard blocker, which is good bc #332 will be done later than originally intended. Moving to m2.

jezdez:
Now that #487 is merged this is fixed.

PR #481, opened 2017-05-19
Use EMR datetimes for status updates.

jezdez:
Fix issue #264.


sunahsuh:
Am I correct that created_at in the cluster model now serves as both the creation date of the django object as well as the cluster? I'm a little troubled by that conflation if so.

atmo/models.py
```diff
@@ -56,14 +58,26 @@ def remove(self):
...
+    created_at = models.DateTimeField(
+        editable=False,
+        blank=True,
+        default=timezone.now,
+    )
+    modified_at = models.DateTimeField(
+        editable=False,
+        blank=True,
+        default=timezone.now,
+    )
```
jezdez: So basically auto_now & auto_now_add suck since they work on model field level and are only ever applied once the object is stored, during the call of the individual field's `pre_save` method. That leaves the object in a weird state where other fields with a default have already set the instance attribute but the auto_now/auto_now_add fields haven't.

A long time ago (in Django 1.0 times) we tried to remove both from Django but failed due to a release schedule conflict. Over the years too many people have expected this weird behavior and now we're stuck with it. The better alternative is to just use the `default` param and override the `save` method for `auto_now`.

But to answer your question, this was triggered by an [issue in factory_boy](https://github.com/FactoryBoy/factory_boy/issues/102) that would prevent the fake object's attributes to be overridden if needed.

jezdez:
@sunahsuh Oh no no, that'd trouble me as well. What we're doing is use the end date from the EMR API to:

a) [set the `finished_at` field](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/models.py#L249-L256) when updating the status when [saving](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/models.py#L272)/[deactivating](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/models.py#L289) a cluster

b) [check on the cluster status periodically](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/tasks.py#L128-L132) and update it if needed

jezdez:
@sunahsuh Ah, I should note that I added returning the `CreationDateTime` and `ReadyDateTime` to the provisioner methods for [cluster info](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/provisioners.py#L81-L83) and [cluster lists](https://github.com/mozilla/telemetry-analysis-service/blob/133327ac2045fadcd2f9551e39d083b0de76d3e9/atmo/clusters/provisioners.py#L119-L121) for the sake of completeness (and since it changes so many tests). Right now we're not actually using that info in the cluster status update code, since I'm not sure what it matters to our users when the cluster was started on EMR or when it was ready to run jobs. If you think I should store that as well in the cluster object, let me know I can extend the PR.

sunahsuh:
I was confused since #264 talks about trying to get cluster CPU hours using the ATMO DB -- if that's a use case we'd like to support then we probably need another field to save the cluster's start time.

atmo/clusters/models.py
```diff
@@ -246,14 +256,26 @@ def provisioner(self):
...
+            ('end_datetime', 'finished_at'),
+        )
+        # set the various model fields to the value the API returned
+        for api_field, model_field in model_field_map:
+            field_value = info.get(api_field)
+            if field_value is None:
+                continue
+            setattr(self, model_field, field_value)
+        if commit:
+            self.save()
```
robhudson: Would it make sense to do the same changes to the atmo/jobs ?

sunahsuh:
Now that the test is passing, LGTM!

atmo/clusters/models.py
jezdez: @robhudson Yeah, probably a good idea.

jezdez:
@robhudson I've filed #487 for the changes relating the jobs app you requested.

PR #482, opened 2017-05-19
Fix #478: Record cluster size to job runs table

### LINKED ISSUE #478: Record cluster size when running jobs for historical log

robhudson:


atmo/jobs/models.py
```diff
@@ -314,6 +315,11 @@ class SparkJobRun(EditedAtModel):
         blank=True,
         null=True,
     )
+    size = models.IntegerField(
+        help_text="Number of computers used to run the job.",
+        blank=True,
+        null=True,
+    )
```
jezdez: Please add `size` to the `autorepr` call that sets `SparkJobRun.__repr__`.

robhudson: Should we throw the EMR version in there as well?

jezdez: Yep, good idea.

Issue #480, opened 2017-05-19
Duplicate job runs on 5/18


sunahsuh:
My txp-install-uninstall-events job ran twice yesterday at 18:04 UTC, along with a few (but not all) other scheduled jobs.

Looking at the celery task results at https://analysis.telemetry.mozilla.org/admin/celery_monitor/taskstate/?name=atmo.jobs.tasks.run_job&tstamp__day=18&tstamp__month=5&tstamp__year=2017 it looks like the tasks that ran twice were once that were retried after a ThrottlingException (click on one of the tasks with an ETA and you can see the exception traceback under Details.) Two questions stemming from this:

- Why did all the jobs run at 18:04? (cc @robotblake -- is there a script running to queue these?)
- Why did the jobs that were retried run twice? (auto-retry logic here: https://github.com/mozilla/telemetry-analysis-service/blob/27d482cdcb2c9974ccece28c3d39140a1f19028b/atmo/celery.py#L69-L86) (My personal hunch is that the ThrottlingException happened after the cluster was started but before we were able to get a job ID back from EMR.)

jezdez:
I think I'll refactor the autoretry a bit to not use a decorator anymore and handle exceptions inside the task function instead.

jezdez:
So #483 moves the exception handling into the task run instead of using Celery's retry mechanism. I think that should give us a few less surprises when it comes to task state changes since the tasks effectively keep running in the same process.

robhudson:
On stage today I tested both launching a new cluster and triggering a spark job. Both ran without duplicates or errors looking at the celery monitoring. (Aside: All the datestamp recording is working as expected as well)

PR #483, opened 2017-05-22
Handle boto ClientError exceptions inline instead of retrying whole tasks.

jezdez:
Refs issue #480.

Also:

- Use the backoff Python library instead of custom code.


jezdez:
@sunahsuh @robhudson So I figured to get better control over the exception handling, I want to step back a bit from using Celery's own task retries and handle the boto ClientError exceptions inline of a Celery task run. The [backoff](https://github.com/litl/backoff) library allows that.

I've also updated the soft timeout and expiry timeouts a notch to better fit the backoff decay.

PR #487, opened 2017-05-23
Use EMR datetimes for status updates of Spark job runs.

jezdez:
Refs #264.


atmo/jobs/models.py
```diff
@@ -280,6 +281,11 @@ def save(self, *args, **kwargs):
         if self.expired_date and self.end_date and self.end_date > timezone.now():
             self.expired_date = None
         super().save(*args, **kwargs)
+        # Remove the cached latest run to this objects will requery it.
```
jezdez: This was needed since the cached_property was storing state when a new spark run was created using factory_boy in the tests.

atmo/jobs/tasks.py
```diff
@@ -200,7 +200,7 @@ def run_job(self, pk, first_run=False):
     spark_job = self.get_spark_job(pk)
 
     # update the cluster status of the latest Spark job run
-    updated = self.update_status(spark_job)
+    updated = self.sync_run(spark_job)
```
robhudson: No need for `commit=True` here?

jezdez: Nope, that's the call to the task method `SparkJobRunTask.sync_run` which exists to be able to test this easier. In it, it calls `SparkJobRun.sync` with `commit=True`.

PR #495, opened 2017-05-24
Fix #477: Record LogUri with job history

### LINKED ISSUE #477: Record log URI in spark job runs
This is to be able to more easily associate logs with a specific run

robhudson:


robhudson:
Closing since the `LogUri` we provide EMR isn't the same as the logs we want to display in ATMO.

PR #509, opened 2017-05-29
Spark job run alerts as one-to-many and partial autoretry revert 

### LINKED ISSUE #496: Create run alert only after last retry fails
Sentry:
https://sentry.prod.mozaws.net/operations/atmo-stage/issues/389740/events/10824516/

Currently an alert record gets created after the first failure. Upon retry and failure another alert tries to get created but fails due to a primary key constraint.

I think it would be preferable if we only created the alert after the last retry and failure so the user gets one email after all attempts have failed. Which is better than getting an email and the 2nd retry worked.

jezdez:
This should at least partially fix #496.

This will also create only an alert if it differs between syncs.

This also partially reverts 58baeb58ce9d49347ee1f65d01ac7424380d2c69 since that moved the retry mechanism in-process and triggered although not caused the issue reported in #496 in the first place.


jezdez:
There is some indentation changes happening here, best is to append `?w=1 `to the URL of this PR to not show it: https://github.com/mozilla/telemetry-analysis-service/pull/509/files?w=1

atmo/jobs/models.py
```diff
@@ -395,24 +395,26 @@ def sync(self, info=None):
...
     A data model to store job run alerts for later processing by an
     async job that sends out emails.
     """
-    run = ForgivingOneToOneField(
+    run = models.ForeignKey(
         SparkJobRun,
         on_delete=models.CASCADE,
-        related_name='alert',  # run.alert & alert.run
+        related_name='alerts',
         primary_key=True,
```
sunahsuh: If this the run id is the primary key for the alert model isn't this still a one-to-one?

atmo/jobs/models.py
```diff
@@ -395,24 +395,26 @@ def sync(self, info=None):
...
-            ))
+            transaction.on_commit(lambda: self.alert(info))
         self.save()
         return self.status
 
+    def alert(self, info):
+        self.alerts.get_or_create(
+            reason_code=info['state_change_reason_code'],
+            reason_message=info['state_change_reason_message'],
+        )
```
sunahsuh: Given that this is expected to return only one object at the most, should there be a `unique_together` on `SparkJobRunAlert` for `(run, reason_code, reason_message)`?

jezdez: Hm, yeah. Let's do that.

atmo/jobs/models.py
jezdez: Hmmm, it's even worse, it's an implicit one-to-one, which is bad. I think the idea here is to be able to have multiple alerts (if their reason differs), so we'll need to change the primary key.

atmo/clusters/tasks.py
```diff
@@ -57,35 +56,35 @@ def send_expiration_mails():
...
-@backoff.on_exception(
-    backoff.expo,
-    ClientError,
-    # This task runs every 5 minutes (300 seconds),
-    # which fits nicely in the backoff decay of 8 tries
-    max_tries=8,
-)
-def update_clusters():
+# This task runs every 5 minutes (300 seconds),
+# which fits nicely in the backoff decay of 8 tries total
```
robhudson: Does the comment still match?

atmo/clusters/tasks.py
```diff
@@ -106,34 +105,40 @@ def update_clusters():
...
+
+                # if not given enqueue a job to update the public IP address
+                # but only if the cluster is running or waiting, so the
+                # API call isn't wasted
+                if (not cluster.master_address and
+                        cluster.most_recent_status in cluster.READY_STATUS_LIST):
+                    transaction.on_commit(
+                        lambda: update_master_address.delay(cluster.id)
+                    )
+        return updated_clusters
```
robhudson: Will other errors be handled by celery itself?

jezdez: Yeah, Celery will handle those on its own and we'll see it in Sentry.

atmo/clusters/tasks.py
jezdez: Yep, still fits.

PR #523, opened 2017-06-01
Fix #460: Add Zeppelin examples to cluster detail

### LINKED ISSUE #460: ATMO should have Zeppelin URL as well as Jupyter
There should be a copy+pastable URL for connecting to Zeppelin on the home page, possibly linking to some docs and instructions for that as well (since it sounds like we are supporting Zeppelin moving forward).

robhudson:
screenshot 2017-06-05 08 24 25

### DIFF
--- a/atmo/templates/atmo/clusters/detail.html
+++ b/atmo/templates/atmo/clusters/detail.html
@@ -48,73 +48,136 @@
       {{ cluster.expires_at }} UTC, <strong>in {{ cluster.expires_at|timeuntil }}</strong>.
     </div>
     <h3>Access</h3>
+    <p>
+    <div class="btn-toolbar" role="toolbar" aria-label="toolbar">
+      <div class="btn-group" data-toggle="btns">
+        <a href="#access-jupyter" class="btn btn-default active" id="access-jupyter-tab" role="tab" data-toggle="tab" aria-controls="access-jupyter">Jupyter</a>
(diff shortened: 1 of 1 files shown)


jezdez:
@robhudson Let's use the same tab style as on the Spark job detail page, the one you used was not working well in responsive mode and is visually noisy.

robhudson:
@jezdez I checked the responsive mode and it looked good to me. I specifically chose this style b/c it matches what is used in the Spark documentation so it should be a common pattern for the user:

screenshot 2017-06-02 08 00 57

jezdez:
@robhudson Please keep the UI consistent with the rest of the ATMO. I understand that the tabs are used elsewhere but that's an issue of Bootstrap in general and we can make a difference here.

The "tabs" included in Bootstrap don't work well on a small screen like on a phone (in responsive mode) because they lack the expressiveness of the button groups. The action done by them is to switch the mode of the whole page, so the UI should express to the user that it requires action.
But the subtle design of Bootstrap's tabs don't convey that message due to the thin outline and no clear separation between its states. The button group on the other hand show that clearly, offering a consistent outline, a clear indicator of which state is active. They also fit nicely in with other action triggering UI elements like other individual buttons on the page.

robhudson:
Updated with buttons.

rafrombrc:
r+

Issue #549, opened 2017-06-15
Better naming for scheduled jobs


mreid-moz:
In #459 identifiers for adhoc clusters and scheduled jobs were changed to be randomly generated. This is fine for a single-use cluster, but makes notifications for scheduled jobs meaningless. 

For example, I received a notification today that job "fervent-snyder-2799" has failed. It is owned by someone else, but emails are cc'd to telemetry-alerts as well (general audience to monitor for important failures).

I would prefer that the name for a scheduled job be reflective of what is being done in the job. 

Alternatively, we could ensure that the "description" field has been filled in and include that information in the notification emails.

PR #576, opened 2017-06-23
added description of spark job to failed run alert email

comloo:
Not sure if I should add the description to the title of the mail as well; I think having it in the body is good enough.

### DIFF
--- a/atmo/templates/atmo/jobs/mails/failed_run_alert.mail
+++ b/atmo/templates/atmo/jobs/mails/failed_run_alert.mail
@@ -7,8 +7,10 @@
 {% block cc %}{{ settings.DEFAULT_FROM_EMAIL }}{% endblock cc %}
 
 {% block body %}
-Your scheduled Spark job "{{ alert.run.spark_job.identifier }}: {{ alert.run.spark_job.description }}" has failed
+Your scheduled Spark job "{{ alert.run.spark_job.identifier }}" has failed
 at approximately {{ alert.run.created_at }} UTC.
+Description: {{ alert.run.spark_job.description }}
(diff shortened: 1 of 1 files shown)


atmo/templates/atmo/jobs/mails/failed_run_alert.mail
```diff
@@ -7,7 +7,7 @@
 {% block cc %}{{ settings.DEFAULT_FROM_EMAIL }}{% endblock cc %}
 
 {% block body %}
-Your scheduled Spark job "{{ alert.run.spark_job.identifier }}" has failed
+Your scheduled Spark job "{{ alert.run.spark_job.identifier }}: {{ alert.run.spark_job.description }}" has failed
```
mreid-moz: I think it would be best to put the job description on the next line - it could be quite long.

mreid-moz:
Note that this addresses #549

mreid-moz:
@comloo I don't think the description needs to go in the email subject, adding it to the email body is sufficient imo

atmo/templates/atmo/jobs/mails/failed_run_alert.mail
```diff
@@ -9,6 +9,8 @@
 {% block body %}
 Your scheduled Spark job "{{ alert.run.spark_job.identifier }}" has failed
 at approximately {{ alert.run.created_at }} UTC.
+Description: {{ alert.run.spark_job.description }}
+
```
robhudson: Can you move this newline above the `Description: ...` so it's separate from the sentence above and only has one line separating the line below?

PR #578, opened 2017-06-27
updated django secret key in docs

comloo:


docs/development.rst
```diff
@@ -24,10 +24,13 @@ To set the application up, please copy the ``.env-dist`` file to one named
...
 Set the ``DJANGO_SECRET_KEY`` variable using the output of the following
-command:
+command, based on which Python your container is using:
 
 .. code-block:: console
+    # Python 2.7
+    python -c "from django.utils.crypto import get_random_string; print(get_random_string(50))"
 
+    # Python 3.6
     python -c "import secrets; print(secrets.token_urlsafe(50))"
```
robhudson: The only thing I dislike about this is it adds extra burden if we upgrade Python version to remember to also update the docs. E.g. if we move to Python 3.7 when it comes out. Also, we shouldn't be using Python 2.7 any longer. 😁 

I'd prefer to suggest that we log in to the Docker container with `make shell` and no matter what version is there things should work.

PR #595, opened 2017-07-05
added ability to upload zeppelin notebook types via atmo

comloo:


robhudson:
There's a few things I'd like to see in addition to what's here:

* Tests!
  * I did a quick scan and didn't see any form tests currently. Would you mind adding some form tests and showing the validation of notebook extension?
  * Maybe a provisioner test with the zeppelin extension to show that the provisioner is called correctly.
* I found an `is_notebook` template tag. That should be updated.

robhudson:
Another thing we'll need a zeppelin equivalent for. Grep for `PUBLIC_NB_URL`. This is used in the job detail page to display the notebook.

maurodoglio:
@comloo one of the tests is still using the old is_notebook filter. Can you please update it?

atmo/jobs/views.py
```diff
@@ -136,6 +136,24 @@ def detail_spark_job(request, id):
 
 @login_required
 @view_permission_required(SparkJob)
+@modified_date
+def detail_zeppelin_job(request, id):
+    """
+    View to show the details for the scheduled Zeppelin job with the given ID.
+    """
+    spark_job = SparkJob.objects.get(pk=id)
```
robhudson: Use `get_object_or_404` here. So if the ID doesn't exist or is invalid it will 404 instead of 500.

atmo/jobs/urls.py
```diff
@@ -14,4 +14,5 @@
     url(r'^(?P<id>\d+)/edit/', views.edit_spark_job, name='jobs-edit'),
     url(r'^(?P<id>\d+)/run/', views.run_spark_job, name='jobs-run'),
     url(r'^(?P<id>\d+)/$', views.detail_spark_job, name='jobs-detail'),
```
robhudson: Do you think we should rename the previous view to match? Like `detail_jupyter_job`?

atmo/jobs/templatetags/notebook.py
```diff
@@ -10,5 +10,12 @@
...
-def is_notebook(value):
+def is_jupyter_notebook(value):
     return value.endswith('.ipynb')
+
+@register.filter
+@stringfilter
+def is_zeppelin_notebook(value):
+    return value.endswith('.json')
+
+
```
robhudson: Remove trailing newlines

atmo/jobs/provisioners.py
```diff
@@ -16,6 +16,11 @@ class SparkJobProvisioner(Provisioner):
 
     def __init__(self):
         super().__init__()
+        # the S3 URI to the zeppelin setup step
+        self.zeppelin_uri = (
+            's3://%s/steps/zeppelin/zeppelin.sh' %
+            constance.config.AWS_SPARK_EMR_BUCKET
+        )
```
robhudson: Since this is in both provisioner subclasses we can move this to the base class.

atmo/jobs/provisioners.py
```diff
@@ -84,6 +89,15 @@ def run(self, user_username, user_email, identifier, emr_release, size,
...
             'Steps': [{
+                'Name': 'setup-zeppelin',
+                'ActionOnFailure': 'TERMINATE_JOB_FLOW',
+                'HadoopJarStep': {
+                    'Jar': self.jar_uri,
+                    'Args': [
+                        self.zeppelin_uri
+                    ]
+                }
+            }, {
```
robhudson: I believe this can go to the base class as well.

atmo/jobs/views.py
```diff
@@ -136,6 +136,24 @@ def detail_spark_job(request, id):
...
+    markdown_url = ''.join([x for x in spark_job.results['data'] if x.endswith('md')])
+    markdown_file = spark_job.provisioner.s3.get_object(Bucket='telemetry-public-analysis-2',
+                                                        Key=markdown_url)
+    response = markdown_file['Body'].read().decode('utf-8')
+    context = {
+        'markdown': response
+    }
+    return TemplateResponse(request, 'atmo/jobs/zeppelin_notebook.html', context=context)
```
robhudson: Write a unit test to test this view.

atmo/jobs/urls.py
comloo: That's a good idea.

comloo: The previous view applies for both notebook types. Zeppelin needs that extra end point because we need a page to render Zeppelin notes whereas a Jupyter note will link to an external renderer.

robhudson: Makes sense, thanks. 👍

PR #600, opened 2017-07-07
Remove unnecessary "depends_on" directives in docker compose config

wlach:
The depends_on directive is needed if one service depends on another,
it is not needed if one service is simply based on another's definition.

https://docs.docker.com/compose/compose-file/#depends_on


wlach:
just something i noticed when examining the configuration to use as a basis for mission control

Issue #225, opened 2017-02-13
Remove ability to specify env vars in .env files


jezdez:
It's a security risk

PR #611, opened 2017-07-13
Move AWS configs from env vars to constance

robhudson:
I saw that jezdez is wanting to move away from env vars, so I'm moving these to constance.


robhudson:
The move to make these configurable is to allow the staging instance of ATMO to use the staging AWS buckets and profiles for EMR bootstrapping, so we can more easily test our bootstrapping scripts.

The move to constance instead of environment variables stems from this:
https://github.com/mozilla/telemetry-analysis-service/issues/225

Thanks for the review!

PR #626, opened 2017-07-19
Counting the things

robhudson:
This is 3 commits. The first adds the table for storing metrics. The next 2 commit implements 3 metrics.


atmo/stats/models.py
```diff
@@ -0,0 +1,55 @@
...
+        editable=False,
+        blank=True,
+        default=timezone.now,
+        db_index=True,
+    )
+    key = models.CharField(
+        max_length=100,
+        db_index=True,
+        help_text='Name of the metric being recorded',
+    )
```
wlach: I think it might make more sense to create a composite index on the key + created_at timestamp instead of creating individual column indexes.

https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field

Or maybe just not bothering with indexing at all.

atmo/stats/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Metric',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_at', models.DateTimeField(blank=True, db_index=True, default=django.utils.timezone.now, editable=False)),
+                ('key', models.CharField(db_index=True, help_text='Name of the metric being recorded', max_length=100)),
+                ('value', models.PositiveIntegerField(help_text='Integer value of the metric')),
+                ('data', django.contrib.postgres.fields.jsonb.JSONField(blank=True, encoder=django.core.serializers.json.DjangoJSONEncoder, help_text='Extra data about this metric', null=True)),
```
wlach: I imagine you're aware of the disadvantages of fields like this in terms of queryability. :)

atmo/stats/models.py
robhudson: I think the plan is to expose this in redash and we'll likely be ordering by `created_at` and grouping by the `key`. So I figured both should be indexed. I'll check out that link tho, thanks.

atmo/stats/migrations/0001_initial.py
robhudson: Yep. :) It may change in the future. The EMR version one that is added here is a test to see how it works out. And is the only metric that requires a string rather than a simple count.

PR #664, opened 2017-08-04
save a redirect

lonnen:
This replaces the wiki link in the footer with the gitbook link the wiki redirects users towards.


lonnen:
think of the perf! a million wasted packets, lost to the ether

PR #684, opened 2017-08-18
Fix #549: remove name generator for jobs

### LINKED ISSUE #549: Better naming for scheduled jobs

washort:


robhudson:
Seems like there isn't a test for this. Could we add one for testing the initial form values on the spark job form?

washort:
test added

Issue #557, opened 2017-06-16
Mount EFS on user's local machine


rafrombrc:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1365284:

If we could get the file system available on user's machines it would make life that much simpler, and possibly decrease the use of clusters by some epsilon amount.

I found some resources in [0] for making it mountable locally (outside of AWS data centers), and [1] for mounting on macOS, probably a majority of our user's machines.

[0] https://serverfault.com/questions/799016/elastic-file-system-efs-mount-outside-of-aws/799213?newreg=dce605ec46b741a28b4e785ca56de8d9
[1] https://serverfault.com/questions/800187/amazon-efs-mount-from-osx

washort:
EFS can be mounted as an NFS volume, via the SSH tunnel already used for the notebook interface.

I plan to add a button on the page that shows the SSH tunnel instructions to show instructions for NFS tunnelling and mounting.

sunahsuh:
> EFS can be mounted as an NFS volume, via the SSH tunnel already used for the notebook interface.

This doesn't address the use case of "I just want to access my data" without spinning up a cluster, which I've definitely run into myself

washort:
Unfortunately this sort of thing is the only option available - EFS is meant for sharing data between EC2 instances only, so tunnelling a connection through one is the only way to access stuff externally. Supposedly, anything meant to be externally accessible should go in S3.

saptarshiguha:
Would it not be possible to have an 'always on' micro instance?

fbertsch:
Hey everyone, we do have an always-on instance: hadoop@ec2-54-202-96-12.us-west-2.compute.amazonaws.com. Your SSH keys are already on there so you should be able to access it.

sunahsuh:
@fbertsch does it have our user directories mounted though? If so, can we 1. add a friendlier dns name and 2. document this somewhere?

fbertsch:
@sunahsuh yes, the volume is mounted at `/home/hadoop/efs`, so for example `/home/hadoop/efs/frank@mozilla.com`. This brings up the fact that these directories are not private and accessible by any ATMO user, but that was true before this machine existed.

mreid-moz:
I don't think they need to be private, but possibly the docs should be updated to make that clear.

fbertsch:
[...]

PR #694, opened 2017-08-25
Instructions for mounting EFS locally (re #557)

washort:
Not sure if these needs tests since it's chiefly a doc change.


robhudson:
@washort Is there a clusters.js file that's missing here?

washort:
Yikes! Rookie mistake. :)

PR #751, opened 2017-09-22
switch from google login to auth0

washort:


atmo/settings.py
```diff
@@ -367,39 +367,17 @@ class Core(AWS, Celery, Constance, CSP, Configuration):
...
-
-    SOCIALACCOUNT_PROVIDERS = {
-        'google': {
-            'HOSTED_DOMAIN': 'mozilla.com',
-            'AUTH_PARAMS': {
-                'prompt': 'select_account',
-            }
-        }
-    }
+    LOGIN_REDIRECT_URL_FAILURE = "/BOGUS"
```
jezdez: If this is really used somewhere, let's just use `'/'` here

atmo/settings.py
```diff
@@ -531,6 +509,20 @@ class Base(Core):
...
+    OIDC_RP_CLIENT_SECRET = values.SecretValue(
+        environ_name='OIDC_RP_CLIENT_SECRET', environ_prefix=None)
+    OIDC_OP_AUTHORIZATION_ENDPOINT = values.Value(
+        environ_name='OIDC_OP_AUTHORIZATION_ENDPOINT', environ_prefix=None)
+    OIDC_OP_TOKEN_ENDPOINT = values.Value(
+        environ_name='OIDC_OP_TOKEN_ENDPOINT', environ_prefix=None)
+    OIDC_OP_USER_ENDPOINT = values.Value(
+        environ_name='OIDC_OP_USER_ENDPOINT', environ_prefix=None)
+    OIDC_OP_DOMAIN = values.Value(
+        environ_name='OIDC_OP_DOMAIN', environ_prefix=None)
```
jezdez: How will this work for local development?

jezdez: What I mean is, will those values be required for local development and if yes, where do I get them? Mind documenting those?

atmo/settings.py
```diff
@@ -566,7 +558,7 @@ def LOGGING(self):
             },
             'loggers': {
                 'root': {
-                    'level': 'INFO',
+                    'level': 'DEBUG',
```
jezdez: Is the root logger required to be DEBUG level?

requirements/build.txt
```diff
@@ -309,3 +299,16 @@ docutils==0.14 \
...
     --hash=sha256:02aec4bd92ab067f6ff27a38a38a41173bf01bed8f89157768c1573f53e474a6 \
     --hash=sha256:51e64ef2ebfb29cae1faa133b3710143496eca21c530f3f71424d77687764274
+mozilla-django-oidc==0.3.1 \
+    --hash=sha256:f906d4ea1879ee588da4a4369013ade19101624516b977fc96b271144123b446 \
+    --hash=sha256:59abd3941ba3471a7baef18a0f704868821cc48ccedd6469fa9e1d08fdc737f3
+python-jose==1.3.2 \
+    --hash=sha256:35eca894ab91db1774251296949679396c46d6bc506ea03804b8f7a7d0204392 \
+    --hash=sha256:968254f57ccd0fc99ab9557f82b90a5b23ccaf3716e8817d8edaa9f21c21bb2d
+pycrypto==2.6.1 \
+    --hash=sha256:f2ce1e989b272cfcb677616763e0a2e7ec659effa67a88aa92b3a65528f60a3c
```
jezdez: Not a huge fan of including pycrypto here since it has been unmaintained for a while now. Is there no way to use cryptography instead?

atmo/settings.py
washort: This was just for testing. It should redirect to "/" but it also does that on success.

atmo/settings.py
washort: I'm unclear on that. This is the documentation I've seen: https://mana.mozilla.org/wiki/display/SECURITY/SSO+Request+Form

requirements/build.txt
washort: Looks like `python-jose` depends on either `pycrypto` or https://pypi.python.org/pypi/pycryptodome .

washort: Oops, upon further review looks like `python-jose` refuses to build without `pycrypto` unless you're using PyPy.

atmo/settings.py
jezdez: Hmm, okay, assuming we're still building a somewhat open project we should add some information how to get this running, at least for our team. Maybe we can find the info on the Auth0 page? I'm sure there are also values that we can provide for people that would like to hook it up directly with Google's OIDC endpoint.

requirements/build.txt
jezdez: Dang it, this sucks, but I guess we can't help it, unless we get python-jose moved to cryptography. Having to maintain a project with pycrypto is a big risk IMO.

jezdez:
@washort One thing I noted while going through the mozilla-django-oidc docs is that we should provide a custom function to use the username portion of the Mozilla email address as the username when creating new users as [specified in the docs](http://mozilla-django-oidc.readthedocs.io/en/0.3.1/installation.html#generating-usernames).

requirements/build.txt
jezdez: Seems like we're not the only ones: https://github.com/capless/warrant/issues/31, https://github.com/mpdavis/python-jose/issues/41 and most notably: https://blog.sqreen.io/stop-using-pycrypto-use-pycryptodome/

jezdez:
Closing in favor of #821.

Issue #528, opened 2017-06-02
Count all the things


robhudson:
Instrument the code (either w statsd calls that we can send to Datadog or log items that will land in ES/Kibana) so we can generate dashboards and possibly notifications.

List of possible metrics (area of concern: CLUSTER = on-demand clusters, SPARKJOB = scheduled Spark jobs):

- [ ] CLUSTER Normalized instance hours of clusters (time between creation and finish multiplied by cluster size)
- [ ] CLUSTER Number of on-demand clusters spun up successfully (to see trends in usage)
- [ ] CLUSTER Number of cluster lifetime extensions
- [ ] CLUSTER / SPARKJOB Number of AWS API error responses and which kind (e.g. throttling exception)
- [ ] CLUSTER / SPARKJOB Number of Python errors/exceptions via Sentry (to see code regressions)
- [ ] CLUSTER / SPARKJOB Number of bootstrapping failures during cluster start up (to track issues with EMR bootstrap script)
- [ ] CLUSTER / SPARKJOB Time between cluster creation (for both scheduled Spark jobs and on-demand clusters) and its readiness to process the first step (the "bootstrapping time" from the user perspective)
- [ ] CLUSTER / SPARKJOB EMR version used for cluster
- [ ] SPARKJOB the time between the cluster's readiness to process the first step and the time when the cluster is shudown (the "runtime of the notebook code" from the user perspective)

robhudson:
@sunahsuh Since we can't video conference, think you can provide details on statsd/datadog vs ES/Kibana here? That way I could get started on instrumenting some of this here and there.

robhudson:
CLUSTER / SPARKJOB EMR version used for cluster is already being recorded with the cluster and spark job runs.

robhudson:
CLUSTER Number of cluster lifetime extensions -- this is being recorded as an increasing integer on the cluster for the total number of extensions.

robhudson:
For the Python errors, we discussed this in IRC:
```
•jezdez> so raven has processors: https://github.com/getsentry/raven-python/blob/master/raven/processors.py
13:49 <•jezdez> which are called whenever an error happens
13:49 <•jezdez> I think we could have one that listens for botocore exceptions and we can record them
13:50 <•jezdez> you can configure the sentry client with the processors to use our custom processor
13:50 <•jezdez> https://docs.sentry.io/clients/python/advanced/#client-arguments
13:52 <•jezdez> that would work for both celery and wsgi
[...]

PR #758, opened 2017-09-26
Issue #528: Record SparkJob metrics

robhudson:


atmo/jobs/models.py
```diff
@@ -406,17 +407,54 @@ def sync(self, info=None):
...
+            if self.status == Cluster.STATUS_TERMINATED_WITH_ERRORS:
+                transaction.on_commit(lambda: self.alert(info))
+            self.save()
+
+        if date_fields_updated:
+            if self.finished_at:
+                # When job is finished, record normalized instance hours.
+                hours = math.ceil(
+                    (self.finished_at - self.started_at).seconds / 60 / 60
+                )
```
jezdez: This looks like a copy/paste of the one in clusters app, can we refactor this into one method please?

atmo/jobs/models.py
```diff
@@ -406,17 +407,54 @@ def sync(self, info=None):
...
+        date_fields_updated = False
+
         # set the various model fields to the value the API returned
         for api_field, model_field in model_field_map:
             field_value = info.get(api_field)
-            if field_value is None:
+            if field_value is None or field_value == getattr(self, model_field):
                 continue
             setattr(self, model_field, field_value)
+            save_needed = True
```
jezdez: This is basically a rewrite of the ["update_fields" feature](https://docs.djangoproject.com/en/1.11/ref/models/instances/#specifying-which-fields-to-save) in Django, let's use this instead, no need to track the state ourselves.

atmo/jobs/models.py
```diff
@@ -406,17 +407,54 @@ def sync(self, info=None):
...
+                # scheduled job to run.
+                run_time = (self.finished_at - self.ready_at).seconds
+                Metric.record(
+                    'sparkjob-run-time', run_time,
+                    data={
+                        'identifier': self.spark_job.identifier,
+                        'size': self.size,
+                        'jobflow_id': self.jobflow_id,
+                    }
+                )
```
jezdez: The record methods uses the model manager `create` method, which means it won't be using an own transaction block and could cause a race condition. Let's wrap this either into an `transaction.on_commit` call or the record method in a `transaction.atomic()` block. The latter is easier since it'll apply to all calls to `Metric.record` but it'll require careful consideration with the app code where it's called from to prevent creating a metric before the app code has been processed. So a `transaction.on_commit` is preferable.

tests/jobs/test_models.py
```diff
@@ -146,7 +147,7 @@ def test_sync_terminated(request, mocker, sparkjob_provisioner_mocks,
         'atmo.clusters.provisioners.ClusterProvisioner.info',
         return_value={
             'creation_datetime': one_hour_ago,
-            'ready_datetime': None,
+            'ready_datetime': one_hour_ago,
```
jezdez: Huh, why this change?

tests/jobs/test_models.py
```diff
@@ -160,6 +161,21 @@ def test_sync_terminated(request, mocker, sparkjob_provisioner_mocks,
     assert spark_job.latest_run.finished_at == now
     assert not spark_job.latest_run.alerts.exists()
 
+    metric = Metric.objects.get(key='sparkjob-normalized-instance-hours')
```
jezdez: Let's check here that there is exactly one `Metric` with that name and not use `get` to fetch it.

tests/jobs/test_models.py
```diff
@@ -160,6 +161,21 @@ def test_sync_terminated(request, mocker, sparkjob_provisioner_mocks,
...
+    assert metric.value == 5
+    assert metric.data == {
+        'identifier': spark_job.identifier,
+        'size': spark_job.size,
+        'jobflow_id': spark_job.latest_run.jobflow_id,
+    }
+    metric = Metric.objects.get(key='sparkjob-run-time')
```
jezdez: Same as above.

tests/jobs/test_models.py
```diff
@@ -251,6 +267,12 @@ def factory():
     return factory
 
 
+def test_is_public(spark_job):
+    assert not spark_job.is_public
+    spark_job.result_visibility = models.SparkJob.RESULT_PUBLIC
+    assert spark_job.is_public
```
jezdez: Yay, coverage!

atmo/jobs/models.py
robhudson: What I'm trying to do is avoid a `save()` if the data didn't change. I noticed that calling `save()` on every sync call made the front-end trigger a UI refresh notice when no actual change actually occurs.

Does using `update_fields` bypass the auto-now stuff for the modified column?

atmo/jobs/models.py
robhudson: Yeah, true. I actually typed it by using the cluster one as reference. I should have noticed myself how exact they were.

tests/jobs/test_models.py
robhudson: This is for the test for recording the "sparkjob-run-time" which calculates run time by using `end_datetime` minus `ready_datetime`. Also, it doesn't make sense to have a end date w/o a ready date.

tests/jobs/test_models.py
robhudson: Curious why? The `get` would fail if there are not exactly 1.

tests/jobs/test_models.py
robhudson: I was checking the coverage on jobs/models.py and noticed this was an easy one to throw in. :)

robhudson:
Updated to wrap metrics recording in transactions.

Any further refactoring will be done in later commits.

robhudson:
Tested locally with clusters and sparkjobs and looks good.

tests/jobs/test_models.py
jezdez: Got it.

tests/jobs/test_models.py
jezdez: Right, it's just a nicer assertion than having to decrypt why it would be returning a `MultipleObjectsReturned` exception

tests/jobs/test_models.py
jezdez: Cool!

PR #812, opened 2017-10-23
Fix #802: Add checking around metrics

### LINKED ISSUE #802: When a job terminates early a metrics recording error is thrown
There might be more than one issue here...

Recently a few jobs terminated early b/c of EBS volume limits. When that happened the status failed to update correctly and an error was reported to sentry when we try to calculate one of the metrics which is missing the `ready_at` timestamp.

If a job terminates with errors I think it best to not attempt to record the usual metrics.

robhudson:
This code incorrectly assumed that `ready_at` will exist if
`finished_at` exists, which is not true when we have a bootstrap
termination with errors.

### DIFF
--- a/atmo/jobs/models.py
+++ b/atmo/jobs/models.py
@@ -461,8 +461,10 @@ class SparkJobRun(EditedAtModel):
                         }
                     )
 
+                if self.finished_at and self.ready_at:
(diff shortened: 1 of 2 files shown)


jezdez:
@robhudson r+

PR #821, opened 2017-10-24
Switch from google login to auth0

jezdez:
Replaces #751.


jezdez:
so apparently the oidc library has a RefreshIDToken middleware that is conflicting  with the test client, I've had to disable it for testing since we're using force login to run the tests

jezdez:
@robhudson @washort Okay, ready to test for real now.

jezdez:
@washort Are you planning on taking a look at the stuff I added on top of your changes, too?

PR #1216, opened 2018-07-10
[Bug 1473089] Migrate telemetry-analysis-service to circle2.0

haroldwoo:


haroldwoo:
@jezdez can you make sure the tests look okay on circle v2? 

https://circleci.com/gh/mozilla/telemetry-analysis-service/3297?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+      - run: docker --version
+      - run: ./bin/test
+
+  deploy:
+    machine:
+      enable: true
+    working_directory: ~/mozilla/telemetry-analysis-service
+    steps:
+      - checkout
+      - run: ./bin/build && ./bin/deploy
```
jezdez: Any specific reason why this is called in one line instead of in two separate ones?

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+    jobs:
+      - build:
+          filters:
+            tags:
+              only: /.*/
+
+      - test:
+          filters:
+            tags:
+              only: /.*/
```
jezdez: I think in the old version we ran build before the test as well, to make sure we didn't break the Docker setup in PRs. Does that do that here as well?

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - deploy:
+          requires:
+            - test
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              only: master
```
jezdez: Does this filter mean it happens for tags **and** only in master?

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+# DOCKER_USER
+# DOCKER_PASS
+#
+
+version: 2
+jobs:
+  build:
+    machine:
+      enable: true
+    working_directory: ~/mozilla/telemetry-analysis-service
```
jezdez: I think you can abstract the working_directory away in a global config if all jobs have the same.

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - checkout
```
jezdez: There is no need to setup docker?

jezdez: The same for the jobs below.

.circleci/config.yml
haroldwoo: I read somewhere that runs run in separate subshells

.circleci/config.yml
haroldwoo: in this case, test and build will run as separate circleci builds.  If either fail, then we wont be able to merge the PR

.circleci/config.yml
haroldwoo: for all tags, on only the master branch.  Is this not what we want?

.circleci/config.yml
haroldwoo: no. docker is pre installed

.circleci/config.yml
haroldwoo: without this, then the deploy build runs even on PR commits

jezdez: Gotcha, thanks for clarification!

.circleci/config.yml
jezdez: Cool!

.circleci/config.yml
jezdez: Cool!

.circleci/config.yml
jezdez: Ah, in parallel?

haroldwoo: ah no. I think similar to how Dockerfiles construct images in layers, separate but sequential.

PR #1257, opened 2018-08-15
Support SameSite cookie attribute on sessionid.

### LINKED ISSUE #1128: Support SameSite cookie attribute on sessionid
Firefox 60 introduces support for the SameSite cookie attribute: https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/
This provides significant protection against CSRF vulnerabilities and so it should be applied to the sessionid cookie.
It looks like its been added to Django https://code.djangoproject.com/ticket/27863 but the fix doesnt appear the be in the latest release or on older streams.
We have various other services that are also on django 1.11.x so I'll look into whether we can get the fix back-ported to that stream.

jezdez:
Fix #1128.

### DIFF
--- a/atmo/settings.py
+++ b/atmo/settings.py
@@ -339,6 +339,7 @@ class Core(AWS, Celery, Constance, CSP, Configuration):
     ]
 
     MIDDLEWARE = (
+        'django_cookies_samesite.middleware.CookiesSameSite',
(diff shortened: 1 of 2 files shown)


robhudson:
Aside from the test error this looks good to me.

